PR #14, opened 2026-07-03
superseded: old stamina OCR consensus

CodeLtDave:
Superseded by #26 after the branch was renamed and the change was rebuilt on current main.

The old 5-sample stamina OCR consensus approach is no longer the active PR direction.


CodeLtDave:
I noticed some general issues with the ocr, I will try to fix them in another way. Rereading the value might solve the problem, but its cleaner to fix the problem on the ocr side instead.

PR #16, opened 2026-07-04
fix(ocr): soft grayscale + bilinear preprocessing to stop digit misreads

CodeLtDave:
### 📖 Summary

Region OCR occasionally returns a **valid-format but wrong** value — a digit flipped to a plausible neighbour (`2→7`, `5→3`). Because the result is a well-formed time/fraction, it **passes validation and is used as-is, with no retry**, so it silently corrupts whatever consumes it: a gather march recorded as returning in **7h instead of 2h** leaves its queue idle for hours; a chest/travel timer read wrong wastes task-queue time; a stamina fraction misread mis-gates stamina-limited tasks.

Observed live across several routines:

| Region | Real value | OCR read |
|:--|:--|:--|
| Gather march timer | `02:28:52` | `07:28:52` |
| Gather march timer | `04:12:10` | `04:17:10` |
| Chief stamina | `215/200` | `213/200` |
| Storehouse chest timer | `01:59:59` | `01:39:39` |

### 🔬 What the investigation found

Captured the raw region crops feeding OCR and inspected them:

- The **input is clean** — the crops are perfectly legible (e.g. an unmistakable `02:28:52`). So it is **not** a region, font, or "can't read" problem.
- The misread is a **recognition-level flip**, and it is **unstable**: the same glyph reads correctly on most frames and flips on others (in one scan one march read correctly while another flipped).
- Root cause is the **preprocessing**. `cropAndPreprocess` hard-binarised each pixel (within `CHANNEL_TOLERANCE` of the target text colour → black, else white) and blew it up by **nearest-neighbour**. The resulting blocky, stair-stepped glyph edges are unlike the anti-aliased text Tesseract's LSTM was trained on, and push borderline digits over the model's decision boundary. (The file-based OCR path already used bilinear smoothing — only the live path was nearest-neighbour + hard-binary.)

### ✨ Change

`cropAndPreprocess` now builds a **soft distance-to-text-colour grayscale** (anti-aliased) at native resolution and upscales it **bilinearly** — keeping the background-removal benefit while giving Tesseract the smooth edges its model expects. One method; no API, config, or behaviour changes elsewhere.

### 📊 Verified offline against real captures

Replayed **2461 captured region crops** through the old and new pipelines (the old pipeline reproduced the live reads → faithful A/B):

- Clean reads on the timer set: **71 → 121**.
- **5 / 5** manually-verified value-flips corrected (the four above + one more).
- Only downside: **1 format-break** on a tiny fast-ticking timer (`00:20:11 → 00:2011`) — which the validator **rejects** (a safe retry), never a silent wrong value.
- No regression elsewhere: status-word regions match case-insensitively; empty / noise regions are unaffected.

### 🛡️ Behaviour / risk

- Pure preprocessing change on the OCR input — no config, UI, or scheduling changes.
- Trades rare **silent-wrong** reads for even-rarer **loud (validator-caught)** reads on two already-unreliable micro-timers. Loud beats silent.

### ✅ Testing

- `mvn -o compile` → `BUILD SUCCESS`.
- Offline A/B harness over the 2461-crop corpus (numbers above); the flip crops re-OCR correctly under the new pipeline.

### DIFF
--- a/fg-vision/src/main/java/dev/frostguard/vision/ocr/TesseractOcrProvider.java
+++ b/fg-vision/src/main/java/dev/frostguard/vision/ocr/TesseractOcrProvider.java
@@ -268,9 +268,16 @@ public final class TesseractOcrProvider {
     // =====================================================================
 
     /**
-     * Extracts a rectangular region, magnifies by nearest-neighbour, and
-     * optionally binarises by proximity to a target text colour.
-     * All three passes are fused into a single traversal.
+     * Extracts a rectangular region and magnifies it for OCR. When isolating
+     * text ({@code stripBackground} + {@code textColour}), the region is turned
+     * into a <em>soft</em> distance-to-target grayscale (anti-aliased) rather
+     * than a hard black/white mask, then upscaled <em>bilinearly</em>.
+     *
+     * <p>The previous approach — hard per-pixel binarisation blown up by
+     * nearest-neighbour — produced blocky, stair-stepped glyph edges that made
+     * the LSTM flip borderline digits on clean input (verified: 2->7, 5->3).
+     * A smooth grayscale keeps the "background removed" benefit while giving
+     * Tesseract the anti-aliased edges its model was trained on.
      */
     private static BufferedImage cropAndPreprocess(RawImageData capture,
             int cx, int cy, int cw, int ch, int scale,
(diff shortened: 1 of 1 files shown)


CodeLtDave:
1783085945782_07-11-09_debug

PR #31, opened 2026-07-13
refactor(build): centralize deps, fix test execution, remove dead weight

Shederator:
Build & dependency management
- Consolidate all dependency and Maven-plugin versions into the parent POM via <dependencyManagement>/<pluginManagement>; child modules now inherit versions (single source of truth). Removes version skew, notably the logback 1.4.14 vs 1.5.13 conflict across parent/fg-app/fg-watcher.
- De-duplicate inline versions (jackson 2.17.2 x3, tess4j 5.14.0 x2, JUnit, plugin versions) previously scattered across 6 POMs.

Test infrastructure (critical)
- Pin maven-surefire-plugin 3.2.5 so the existing JUnit 5 tests are actually discovered and run. Previously 'Tests run: 0' — the two regression test classes (GatherQueuePolicyTest, StaminaServiceTest) never executed. Now all 6 tests run and pass on every build.

Dead code / dependency hygiene
- Remove unused JDA (Discord) dependency and its transitive Netty stack (~14 runtime jars). The only Discord references are a UI invite link/icon; the actual bot init call was commented out. Cuts the packaged runtime lib from 90 to 76 jars.
- Fix a latent hidden transitive dependency: fg-tasks/fg-app used org.jetbrains.annotations.@NotNull that was leaking in through JDA. Declare org.jetbrains:annotations explicitly at 'provided' scope (compile-time only, no runtime bloat).

Repository size / binaries
- Delete 102 MB of dead, duplicated native binaries: lib/opencv and tools/opencv each held an unused 51 MB copy of opencv_java4110.dll. The canonical DLL loaded at runtime lives in fg-vision resources and is extracted from the classpath on launch.
- Track *.dll via Git LFS (previously only *.exe/*.traineddata were), so the remaining native binaries no longer bloat regular Git history.

Performance / robustness
- OpenCvPatternLocator.extractAndLoadNative is now idempotent: skips the ~50 MB re-extraction when an up-to-date copy exists, uses NIO buffered copy, and writes+atomically-moves to avoid truncated/corrupt binaries.

Developer experience
- Add GitHub Actions CI pipeline (JDK 21 + Git LFS) running mvn clean install with test reports and desktop-bundle artifacts.
- Remove spurious Node.js scaffolding (package.json/package-lock.json) that had a broken description and did not belong in a Java/Maven project.
- Rewrite .gitignore to match the actual Java/Maven layout (drop dead fg-web references; ignore runtime-extracted lib/opencv).


CodeLtDave:
I tested this PR locally in the existing PR worktree and also did a real emulator smoke run.

Validated on commit `74545a6`:

- `mvn -pl fg-app -am package -DskipTests=false` completed successfully.
- Surefire now executes the existing JUnit 5 tests: 6 tests, 0 failures.
- The desktop bundle was produced successfully: `fg-app/target/frostguard-2.1.0-desktop-bundle.zip`.
- Runtime `fg-app/target/lib` contains 76 jars, with no JDA/Netty jars present.
- `git lfs fsck` passed.
- LFS-managed native files checked out as real binaries, not pointer files.
- The packaged ADB binary runs successfully (`adb version`).
- The built app starts from `fg-app/target` with `java --enable-native-access=ALL-UNNAMED -jar frostguard-2.1.0.jar`.
- OpenCV native loading is confirmed in the log: `Native library loaded from: lib\opencv\opencv_java4110.dll`.

Real emulator smoke evidence:

- MuMu emulator was launched/connected through the built app.
- Whiteout Survival was launched.
- Home screen detection succeeded.
- Initial stamina OCR succeeded (`295/200`, parsed as `295`).
- A Training routine started and performed real template/OCR/ADB interactions against the emulator.
- I checked the app and account logs for startup/native/classpath failures: no `ERROR`, `UnsatisfiedLinkError`, `NoClassDefFoundError`, or `ClassNotFoundException` entries found.

There was one routine-level warning during Training (`Train button not detected`) after normal in-game navigation/template checks, but that is not related to this PR's build/dependency/LFS/native-loading scope.

Based on this validation, I consider PR #31 safe to merge.

CodeLtDave (approved):

PR #36, opened 2026-07-14
feat(march): estimate queue wakeups from slot state

CodeLtDave:
> **Stack 2/3** - Depends on #34. Review and merge order: #34 -> #36 -> #37.

## Summary

Stacked on #34. Please review/merge #34 first; until then GitHub will show the base March Queue model changes in this PR diff too.

This follow-up adds the first shared planning helper on top of the richer March Queue state and moves Polar's no-slot retry planning to it.

## Change

- Add `MarchSlotAvailabilityEstimator` to convert typed March Queue rows into conservative wake-up hints.
- Keep Returning as the only exact slot-release countdown.
- Treat Gather, Attack, and Rally timers as lower-bound recheck hints instead of slot-free times.
- Add a gather-return buffer for working Gather rows because the visible timer is gather work remaining, not home arrival.
- Back off longer for stationed rows with no automatic release timer.
- Move Polar no-slot retry planning onto the estimator while keeping routine-launched rally returns as exact first-class evidence.

## Commit Split

- `2bdd0f1 feat(march): estimate queue wakeups from slot state` adds the generic estimator and tests.
- `277fdca feat(polar): use march queue wakeup estimates` migrates Polar as the first consumer.

## Tests

- `mvn -pl fg-engine -am test -DskipTests=false` -> 38 tests, `BUILD SUCCESS`.
- `mvn -pl fg-tasks -am test -DskipTests=false` -> `BUILD SUCCESS`.

Earlier same-content package verification passed with:

- `mvn -pl fg-app -am package -DskipTests`

## Next

After #34 and this follow-up are settled, Gather can use the same slot state for resource coverage, no-slot wakeups, and gather-specific deployment decisions.


Shederator:
Merged into `main` in commit 9f5fb0e ("Merge PR #36 …").

The two commits (estimator + Polar migration) were cherry-picked onto the current `main` tip rather than merging this branch directly, because the branch base predates PRs #28/#32/#33/#35. Merging the branch as-is would have shown those already-merged features as reverts in the diff. Cherry-picking keeps the landed diff to just the 5 files this PR actually changes. Original authorship (David Erdös) preserved.

Verified locally: full `mvn install` compile + 33 unit tests pass (MarchSlotAvailabilityEstimatorTest 5/5, MarchSlotStateCompatibilityTest 2/2). The only skipped test is the OpenCV-native evidence test, which needs the native OpenCV lib not present in the sandbox. Closing as merged.

PR #62, opened 2026-07-28
ci: add daily Windows bundle build

bizulk:
Push a daily build, so that users can at least download artifact from those build and provide feedback.
I've checked that we can cross-build a package from Linux.
[Issue description ](https://github.com/bizulk/wosbot/issues/1)


bizulk:
This CI is not perfect as it skip tests, we can't run them until we resolve the mult-platform support issue
You can check an action example [here](https://github.com/bizulk/wosbot/actions/runs/30400388697).

Shederator:
Updated PR #62 with fixes to make the workflow production-ready:
Fixed a pre-existing packaging issue where the ZIP was created before dependencies were copied, resulting in empty bundles on clean builds.
Made OpenCV native loading platform-aware, so all 97 tests now pass on Linux instead of being skipped.
Added LFS pointer validation and manifest cross-checks to catch broken release artifacts.
Also added PR trigger, job timeout, version-tagged artifact uploads, Surefire reports, refreshed CI docs, and removed the obsolete workflow file.
The workflow has been verified on JDK 21/Maven and should work automatically after merge with no additional setup. Testing needed.

Shederator:
Pushed a follow-up commit. The workflow built the right artifact, but several of its guards could not actually fail, so a broken release would still have gone out green. Full pipeline re-verified from a pristine clone on JDK 21.

### Confirmed the packaging fix is real

Built current `main` in isolation: the ZIP is **17 MB with 0 runtime JARs** and no JavaFX at all. `copy-dependencies` and the assembly were both bound to `package`, where Maven honours POM declaration order, so the ZIP was assembled before `target/lib` existed. It only ever looked right when a previous build had left `target/lib` behind — never true on a clean runner. With `prepare-package` a clean build yields **218 MB / 76 JARs**.

### False-pass guards that are now fixed

| Guard | Problem | Now |
|---|---|---|
| Manifest cross-check | Substring match — `lib/foo.jar` was satisfied by an unrelated `lib/foo.jar.disabled` | Exact comparison against the ZIP central directory |
| `require 'templates/'` | Satisfied by the bare directory entry with every sprite missing (same for `custom_tasks/`) | Requires actual `.png` / file entries |
| LFS materialisation | `git lfs ls-files` returning nothing made the loop, and the step, pass silently | Fails on empty list, asserts the 4 critical assets are tracked, adds a size floor |
| Missing checks | `AdbWinUsbApi.dll`, `osd`/`chi_sim` models and docs were never verified | Explicitly required |
| JavaFX leak check | Only rejected `-linux` | Rejects any foreign classifier |

### Two verification layers

`verify_bundle.py` proves the right files are at the right paths; it cannot prove they link together. A dependency dropped from a POM yields a structurally perfect bundle that dies with `NoClassDefFoundError` on first launch, so `smoke_test_bundle.sh` resolves the real entry points off the real bundle classpath, checks `java -jar` resolves the manifest `Class-Path`, and boots the shaded watcher for real.

Validated against a deliberately damaged bundle — removing `lib/hibernate-core-*.jar` is caught by the manifest cross-check **and** independently by the smoke test. `ci/test_verify_bundle.py` adds 14 unit tests asserting each guard rejects the regression it exists for, and runs in CI so the verifier itself cannot silently stop working.

### OpenCV loader

A clone without Git LFS leaves a ~130 byte pointer stub in place of the 52 MB DLL, which previously killed bootstrap on Windows. The Windows path now falls back to the openpnp image and logs the `git lfs pull` hint; total failure throws `IOException` with OS/arch and cause rather than surfacing later as an unrelated `UnsatisfiedLinkError`. Both branches were exercised directly.

### Also

PR builds now trigger on `ci/**`, `custom_tasks/**`, `fg-watcher.bat` and `.gitattributes` — all previously changeable with no build running. Version resolves in its own step so a failure cannot produce an untagged artifact. Restored tab indentation in `fg-app/pom.xml`, ignored `__pycache__`, and documented the nightly artifact download in the README.

### Results

- **99 tests pass** on Linux (was 97 and skipped entirely before), including the vision/OCR saved-frame suites
- Bundle: 218 MB, 76 runtime JARs, all 76 manifest `Class-Path` entries resolve, 397 template sprites
- `-win` JavaFX runtime present, no foreign classifier
- Smoke test launches both the app and watcher JARs successfully

One thing worth knowing: LFS traffic. Each run pulls ~77 MB of LFS assets, so a daily build plus PR runs is roughly 3-4 GB/month against the free 1 GB LFS bandwidth quota. Worth watching; if it becomes an issue the schedule can be reduced to weekly, or the assets cached between runs.

PR #67, opened 2026-07-29
ci: publish the nightly Windows bundle to Discord with a working download link

Shederator:
## Summary

The nightly Windows bundle was only reachable as a GitHub Actions artifact, which requires a signed-in GitHub account with access to this repository. Testers in the Discord channel had no way to get a build.

Every run on `main` now republishes a rolling `nightly` prerelease with the bundle attached, and posts its **public, login-free** download URL to the webhook held in the `DISCORD_NIGHTLY_WEBHOOK_URL` secret.

Permanent link:

```
https://github.com/Shederator/wosbot/releases/download/nightly/frostguard-windows-desktop-bundle.zip
```

## Why not attach the ZIP to the message

The bundle is ~220 MB. A Discord webhook may upload at most 8 MiB, so the file itself can never be attached; the notifier falls back to a link automatically and only attaches when a file is genuinely small enough.

## Getting the download link right

This is the part that quietly 404s if done naively, so each guard is deliberate:

- **Fixed asset name.** A download URL embeds the asset filename. Uploading `frostguard-2.1.0-desktop-bundle.zip` would change the link at the next version bump and break every message already posted to the channel. The versioned ZIP is copied to `frostguard-windows-desktop-bundle.zip` before upload; the version still appears in the release title, notes and the Discord card.
- **Delete and recreate, not edit.** `gh release edit --target` does *not* move an existing tag. Editing in place would leave `nightly` pinned to the first commit it was ever cut from while the notes advertised a newer SHA. `--cleanup-tag` removes the tag so the recreated one points at the current build.
- **Asset attached at creation time**, in a single call, so there is never a window where the tag exists but the download 404s.
- **The URL is read back from the API** (`browser_download_url`) and then actually fetched before being advertised. Predicting the URL from a filename is exactly how a dead link reaches the channel; now a missing or unservable asset fails the step instead. The notifier independently drops any non-absolute URL and falls back to the run link.

## Notification behaviour

- **Failures notify too** (`if: always()`), so a broken nightly shows as a red card rather than being silently absent — the failure mode a success-only notifier hides.
- **Pull requests never publish or notify.** A fork PR token cannot read secrets, and republishing `nightly` from unmerged code would hand testers an unreviewed build.
- `continue-on-error: true` keeps a Discord outage from turning a good build red. Delivery retries on 429 (honouring `Retry-After`) and on 5xx.
- **No mass pings.** `allowed_mentions: {parse: []}` is structural, so `@everyone` in a commit subject cannot ping the channel.
- **No shell injection.** The commit message is passed through the environment, never interpolated into `run:`, so `$(...)` in a commit subject cannot execute on the runner while the webhook secret is in scope.
- **The webhook is never logged.** Diagnostics are redacted, since Actions logs are public on a public repository.
- A missing secret warns and passes rather than failing a good build.

## Also fixed

- The job summary and the Discord card now derive from one shared metrics step, so they cannot disagree.
- The install hint is `java -jar frostguard-*.jar`. The assembly ships no launcher `.bat` for the app itself (only `fg-watcher.bat`), so the obvious `frostguard.bat` instruction would have sent every tester into a support question.

## Tests

- `ci/test_discord_notify.py` — 22 self-tests, no network. Covers Discord's length limits, mention suppression, token redaction, rate-limit handling, the malformed-URL fallback, and empty step outputs from an early build failure. These run *before* the 30-minute Maven build so a broken payload fails in seconds. The suite caught two real bugs during development (a missing argument definition and an `int("")` crash on the failure path).
- `ci/test_verify_bundle.py` — 14 existing tests, still green.

Verified against this repository: both success and failure payloads returned **HTTP 204** from the live webhook, the release publishes, the URL resolves via the API, and an **anonymous request with no GitHub token downloads it (HTTP 200)**.

### DIFF
--- a/.github/workflows/daily-windows-bundle.yml
+++ b/.github/workflows/daily-windows-bundle.yml
@@ -18,8 +18,12 @@ on:
     - cron: "17 3 * * *"
   workflow_dispatch:
 
+# `contents: write` is needed to publish the rolling `nightly` release that the
+# Discord message links to. An Actions artifact URL requires a signed-in GitHub
+# account with access to this repository, so it is useless as a "downloadable"
+# link for testers in a Discord channel.
 permissions:
-  contents: read
+  contents: write
 
 concurrency:
   group: daily-windows-bundle-${{ github.ref }}
@@ -128,6 +132,13 @@ jobs:
       - name: Test the bundle verifier
         run: python3 ci/test_verify_bundle.py
 
+      # The notifier runs in a non-blocking step, so a bug in it would show up
+      # as "no message arrived" rather than as a red build. Its own tests run
+      # here, before the 30-minute Maven build, so a broken payload contract
+      # fails in seconds.
+      - name: Test the Discord notifier
+        run: python3 ci/test_discord_notify.py
+
       - name: Resolve project version
         id: version
         run: |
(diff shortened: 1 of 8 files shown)


CodeLtDave:
Good job, this is exactly the solution I had in mind. I checked it again and confirmed that Discord receives the permanent GitHub Release link, not the expiring Actions artifact.

The only thing I would still add is the list of PR/commit titles since the previous nightly. Notifications should also be restricted to main so ci/** branch pushes do not notify Discord.

PR #99, opened 2026-08-03
fix(profile-switch): make shared-emulator handovers safe

pernerch:
## Summary

This PR makes same-emulator profile handovers fail safely instead of treating a failed character switch as a successful initialization.

- Return explicit character-switch outcomes instead of a boolean.
- Keep the task gate closed until initialization and the character switch really succeed.
- If an enabled sibling profile shares the emulator, keep the emulator running, release the slot, pause the failed queue, and retry initialization after 5 minutes.
- If no sibling profile can use the emulator, keep the conservative close/restart behavior.
- Publish active-profile ownership from the scheduler only after successful initialization.
- Keep the launcher title synchronized for all active profiles, with deterministic ordering and reset behavior when the bot stops.
- Add focused scheduler, recovery-policy, architecture, and launcher-title tests.

## Why

Previously, a missing character-switch target could return normally from initialization. The queue then recorded initialization as successful and could continue running ordinary tasks on the wrong character. Simply keeping the emulator open made that failure mode more dangerous.

The updated flow preserves the fast handover benefit for shared emulators while preventing work from crossing the failed initialization boundary. The maintainer follow-up also removes hidden cross-profile state from `DelayedTask`, leaving queue ownership in the scheduler layer.

## Validation

- `mvn package`
- Reactor build: **BUILD SUCCESS**
- Tests: **246 run, 0 failures, 0 errors, 0 skipped**
- Desktop JAR and bundle produced successfully
- GitHub Actions: **passed** for the updated head

## Evidence level

Automated tests only. Before merge, the shared-emulator handover should still be verified live for:

1. successful A -> B handover;
2. missing B target while A or another sibling continues using the open emulator;
3. retry after 5 minutes;
4. no-sibling failure causing the conservative emulator restart path;
5. launcher title changes during handover and after stop.

Thanks @pernerch for identifying the shared-emulator use case and providing the original implementation. Please review the maintainer follow-ups, especially whether the retry and ownership behavior matches what you observed in live operation.


CodeLtDave:
Maintainer follow-up pushed in `bc0bee2`.

The original fast-switch idea is preserved, but the failure path is now guarded: a failed character switch no longer counts as a successful initialization, normal tasks stay blocked, and another enabled profile can take over the still-running shared emulator. Without a usable sibling profile, we retain the conservative close/restart behavior.

Local verification: `mvn package` passed with 244 tests and the desktop bundle was produced. Live shared-emulator validation is still required; the exact scenarios are listed in the updated PR description.

@pernerch, could you please review the maintainer changes and, if possible, confirm the handover/retry behavior against your live setup? We have intentionally kept this work in your PR.

CodeLtDave:
Maintainability follow-up pushed in `cd0dde5`:

- active-profile ownership is now published by `TaskQueue` only after `Initialize` succeeds;
- hidden cross-profile state was removed from `DelayedTask`;
- nearby contributor-history comments and indentation were cleaned up;
- duplicate launcher-title refreshes were removed;
- equal display names now use profile ID as a deterministic tie-breaker;
- architecture and formatter regression coverage was extended.

Validation: `mvn package` passed with 246 tests, 0 failures/errors/skips, and produced the desktop bundle. Live handover verification is still the remaining merge gate.

@pernerch, this is the final code-structure cleanup I would especially like you to review alongside the handover behavior.

PR #108, opened 2026-08-03
feat(gather): support level 9 with regional fallback

pernerch:
﻿## Summary
- allow gather level selection up to 9 in HMI for all four resource types
- implement gather deployment fallback chain for region-dependent level availability:
  - configured 9 -> try 9, then 8
  - if "downgrade level if no node is found" is enabled, continue one level lower from the active max attempt
- preserve existing downgrade behavior while enabling 9->7 fallback path when 9/8 are unavailable

## Validation
- `mvn -pl fg-tasks,fg-app -am -DskipTests clean compile` (success)

## Evidence level
- not yet documented in this PR body
- merge readiness requires explicit validation evidence (tests/logs/frames) before approval


pernerch:
Superseded by #123, which carries the same gather level-9 fallback scope on a clean branch based on main with full affected-module tests.

PR #109, opened 2026-08-03
fix(gather): reconcile pool with active marches

pernerch:
Reconciles the persisted gather rotation pool with the current HMI-enabled gather types and the currently active marches at gather startup.

This keeps the HMI as the source of truth while correcting stale pool state after downtime or manual account changes.

Validation:
- `mvn -pl fg-tasks -am -DskipTests compile`


pernerch:
Superseded by #122, which contains the same change on a clean branch based on main with validated module tests.

Issue #95, opened 2026-08-02
fix(config): apply saved changes consistently at runtime


CodeLtDave:
## Problem

Configuration changes made while the bot is running behave inconsistently:

- some changes only take effect after restarting the bot;
- some changes appear to take effect immediately;
- some edited values are reverted while the bot is still running;
- it is unclear whether additional values may be reverted when the bot stops.

At minimum, a successfully saved value must remain persisted and be used after the next bot restart. Ideally, every setting that can safely change at runtime should be applied immediately and consistently.

## Technical context

The current runtime has more than one configuration access pattern:

- `ScheduleService` loads enabled profiles and their configuration when the engine starts;
- `TaskDispatcher` / `TaskQueue` retain the corresponding `AccountDescriptor` instances;
- task routines frequently read profile settings from those in-memory descriptors;
- several global settings are instead loaded again through `ConfigService.loadGlobalSettings()` when used;
- profile editing persists a complete profile/config snapshot asynchronously via `ProfileService.persistAccount(...)`;
- profile data-change broadcasts currently update UI consumers, but the scheduler/queues do not subscribe to replace or refresh their runtime profile state;
- runtime services may also write profile configuration and broadcast another descriptor snapshot.

This explains why global values that are queried repeatedly may change immediately, while profile-scoped values held by running queues remain stale. Full-snapshot persistence and competing UI/runtime updates may also allow an older in-memory snapshot to overwrite or visually revert a newer edit.

## Expected behavior

1. A successfully saved configuration change is never silently overwritten by stale runtime state.
2. After stopping and restarting the bot, the latest saved configuration is used reliably.
3. Settings that support live changes are propagated to the running profile/queue and affect subsequent task decisions.
4. Settings that cannot safely change at runtime are clearly identified in the UI and remain persisted for the next restart.
5. Saving and applying a configuration change is logged with the profile, key, persistence result, and runtime-application result.

## Suggested direction

- Define one authoritative update path for profile-scoped configuration.
- Avoid replacing the entire persisted config set when only one key changed, or add revision/conflict protection to snapshot writes.
- Propagate committed profile changes to the scheduler-owned runtime descriptor/queue.
- Decide per setting whether it is:
  - live-applicable;
  - applied before the next task execution;
  - restart-required.
- Ensure bot shutdown never writes an older cached configuration back to storage.
- Add concurrency/regression tests for rapid edits and simultaneous runtime-owned config updates.

## Acceptance criteria

[...]

PR #122, opened 2026-08-04
fix(gather): apply enabled type changes to rotation

pernerch:
## Summary
- track the gather types that were enabled when the persisted rotation was last updated
- add only genuinely newly enabled types to the current rotation and remove disabled types
- preserve types that are absent because they were already consumed in the current rotation cycle
- initialize existing profiles conservatively without resetting their current cycle

## Scope
- active/manual marches continue to be handled by the existing `fillQueues(...)` logic
- runtime propagation of saved profile changes remains separate and is tracked by #95

## Validation
- `mvn -pl fg-tasks -am test` (pass: 198 tests, 0 failures/errors/skips)

## Evidence level
- automated tests
- live account-log confirmation not performed


CodeLtDave:
Thanks for the PR. I took a closer look at the persisted pool semantics and found that a missing pool entry does not necessarily mean state drift: it can also mean that the type was already used in the current rotation cycle. Re-adding every enabled inactive type would therefore partially reset the rotation whenever a march returned.

I updated the branch to narrow the fix to actual gather configuration changes:

- a persisted enabled-type baseline now identifies genuine config deltas;
- newly enabled types are added to the current pool;
- disabled types are removed;
- already-consumed types remain absent, so the current rotation is preserved;
- active/manual marches remain handled by the existing `fillQueues(...)` path;
- config propagation itself remains out of scope here and is still tracked by #95.

I also rebased the PR onto current `main` and added focused policy tests. `mvn -pl fg-tasks -am test` passes with 198 tests and no failures, errors, or skips.

Could you take a look at the adjusted behavior and confirm that this matches the config-change case you originally wanted to address?

PR #135, opened 2026-08-09
feat(vision): integrate PaddleOCR backend via RapidOCR

Cloudbed-invi:
## Summary
* Abstracted the OCR subsystem into a unified `OcrEngine` facade supporting multiple pluggable providers.
* Integrated RapidOCR (PaddleOCR via ONNX Runtime) as a new OCR backend under `PaddleOcrProvider`.
* Converted all `fg-tasks` (Research, Arena, Rallies) to use the new `OcrEngine` abstraction.
* Added a thread-safe `getProvider()` initialization to support parallel OCR processing via `CompletableFuture`.
* Updated the `SettingsPanel` UI to allow toggling the OCR Engine in real-time.

## Why
Tesseract OCR consistently struggles with game text layouts, misreading numbers and missing smaller buttons completely (e.g. "Cancel", "Confirm"). PaddleOCR offers massively improved accuracy for layout preservation and small text reading, all while maintaining the same ~1.0s processing speed via CPU execution.

## Validation
* **Automated Tests:** 
  - Ran `OcrEngineComparisonTest` locally which parsed 5 identical game screen regions using both Tesseract and PaddleOCR.
  - Verified `target/ocr-benchmark.csv` indicating identical performance speeds but substantially improved accuracy for PaddleOCR.
  - Added and ran `ParallelOcrTest` to ensure that `OcrEngine` can safely handle multiple concurrent requests without crashing the underlying ONNX execution session.
* **Build Verification:** Reactor built successfully (`mvn clean install`) across all modules.

## Risks
* **Dependencies:** `rapidocr-onnx-platform` is now included. Maven handles this automatically, but it expands the final jar size slightly due to bundled ONNX models.
* **Native Memory:** ONNX Runtime runs C++ underneath. I added `synchronized` to the provider initialization to ensure thread safety, but prolonged parallel usage over hundreds of hours could theoretically introduce memory leaks if RapidOCR's JNA garbage collection doesn't properly clean up the `OcrResult` structs. Will monitor in live logs.

### DIFF
--- a/fg-api/src/main/java/dev/frostguard/api/configs/ConfigurationKeyEnum.java
+++ b/fg-api/src/main/java/dev/frostguard/api/configs/ConfigurationKeyEnum.java
@@ -265,6 +265,14 @@ public enum ConfigurationKeyEnum {
     DISCORD_TOKEN_STRING                ("",            String.class,   ConfigCategory.SYSTEM),
     GAME_VERSION_STRING                 ("GLOBAL",      String.class,   ConfigCategory.SYSTEM),
     IDLE_BEHAVIOR_STRING                ("CLOSE_EMULATOR", String.class, ConfigCategory.SYSTEM),
+    /* ─────────── emulators ─────────── */
+
+    ADB_PORT_INT                        ("5555",             Integer.class,       ConfigCategory.SYSTEM),
+    EMULATOR_NAME_STRING                ("",                 String.class,        ConfigCategory.SYSTEM),
+    EMULATOR_PATH_STRING                ("",                 String.class,        ConfigCategory.SYSTEM),
+    OCR_ENGINE_STRING                   ("TESSERACT",        String.class,        ConfigCategory.SYSTEM),
+    INACTIVITY_POLICY_STRING            ("CLOSE_EMULATOR",   String.class,        ConfigCategory.SYSTEM),
+    MAX_CONCURRENT_INSTANCES_INT        ("4",                Integer.class,       ConfigCategory.SYSTEM),
     // Changed by pernerch | Date: 2026-07-04 | Why: allow explicit stop-policy selection for GUI stop action.
     STOP_BEHAVIOR_STRING                ("DO_NOTHING",  String.class,   ConfigCategory.SYSTEM),
     // Changed by pernerch | Date: 2026-07-04 | Why: separate Telegram stop behavior from local GUI stop behavior.
--- a/fg-api/src/main/java/dev/frostguard/api/configs/OcrEngineType.java
+++ b/fg-api/src/main/java/dev/frostguard/api/configs/OcrEngineType.java
@@ -1,0 +1,33 @@
+package dev.frostguard.api.configs;
+
+public enum OcrEngineType {
+    TESSERACT("Tesseract (Default)"),
+    PADDLE_ONNX("PaddleOCR (Experimental)");
+
+    private final String displayName;
+
+    OcrEngineType(String displayName) {
+        this.displayName = displayName;
+    }
+
+    public String getDisplayName() {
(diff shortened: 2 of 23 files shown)


CodeLtDave:
Thanks for working on this — adding a provider-neutral OCR layer and evaluating RapidOCR is a useful direction. However, I don't think this is mergeable in its current form. This is currently closer to a proof of concept than a safe replacement path for the existing OCR behavior.

The main blockers I found are:

- The branch no longer builds against current `main`. `IntelCooldownOcrFrameTest` still calls `TesseractOcrProvider.recognizeText(...)` statically, while this PR makes it an instance method. This is the cause of the current CI failure.
- The Paddle provider silently ignores the language argument and nearly all `TesseractSettingsData` behavior, including page segmentation, whitelists, scaling, thresholding, color/background handling, padding, and inversion. Many current OCR call sites rely on those presets, so switching the global provider changes more than just the underlying engine and can regress working reads.
- The new tests do not currently verify correctness. They have no expected OCR results/assertions, catch failures and turn them into output, and the parallel test references a fixture at a path that does not exist. They can therefore pass even if all Paddle calls fail. They also change global engine state without restoring it.
- The thread-safety claim is not demonstrated. Provider lookup is synchronized, but OCR calls still run concurrently on one native wrapper instance, and `activeType` is mutable global state without a defined switching/lifecycle policy.
- Persisted engine selection appears to be applied by the GUI controller rather than during application bootstrap, so headless behavior is unclear and may remain on Tesseract.
- Cross-platform packaging needs explicit handling. `rapidocr-onnx-platform` selects native dependencies for the build OS, while our Windows bundle is built on Linux. A successful Linux build could therefore still produce a Windows package containing the wrong native library.
- The size increase is substantial rather than slight. The models and native/runtime dependencies add well over 100 MiB in the current dependency graph. The direct Microsoft `onnxruntime` dependency also appears unused by the implementation and should be removed.
- Creating a temporary PNG for every OCR call adds disk I/O to hot paths and needs either a buffer-based implementation or a clearly tested cleanup/lifecycle strategy.
- The additional unrelated, unused config keys should not be part of this PR.

I suggest splitting this into independently reviewable steps:

1. Introduce the provider-neutral interface/factory while preserving Tesseract behavior exactly, with all existing tests green.
2. Add RapidOCR as an opt-in experimental provider, including explicit Windows-native packaging and real saved-frame tests with expected outputs for the OCR presets we actually use.
3. Add configuration/UI switching and initialize it consistently for both GUI and headless startup once the provider behavior is validated.

Before enabling Paddle globally, I would want saved real-frame coverage with assertions, live account-log confirmation across representative tasks, a packaged-Windows native-load smoke test, proper failure/fallback behavior, and measured cold/warm latency, memory use, and final bundle size. Tesseract should remain the default until that evidence exists.

The direction is promising, so I would prefer keeping the PR open as a draft or reworking it in the smaller steps above rather than closing the idea entirely.

Cloudbed-invi:
Closing in favour of the approach suggested in review — splitting into 
independently reviewable steps.

**Step 1** (open): Cloudbed-invi/wosbot → #157 
Provider-neutral `OcrProvider` interface + `OcrEngine` facade. Tesseract 
only, zero behavior change, all existing tests pass.

**Step 2** (in progress):  
RapidOCR/Paddle as opt-in experimental provider. Will include preprocessing 
to map `TesseractSettingsData` (scaling, thresholding, inversion, padding, 
whitelist) onto Paddle so no presets are silently dropped, plus real saved-frame 
tests with exact expected strings.

**Step 3**: Config/UI switching once Step 2 is validated with real logs.

Thanks for the detailed review — the direction was right, it just needed 
to be broken up properly.

PR #157, opened 2026-08-10
refactor(vision): introduce provider-neutral OCR interface (step 1 of 3)

Cloudbed-invi:
## What

Introduces a provider-neutral OCR boundary while keeping Tesseract as the sole provider and default. This is Step 1 of the staged OCR work tracked in #212; it does not add Paddle/RapidOCR or runtime provider selection.

## Why

A second provider should be evaluated behind a stable boundary without changing existing capture, preprocessing, diagnostics, or Tesseract recognition behavior first.

## Changes

- replaces the shared Tesseract-specific settings and exception contracts with `OcrSettingsData`, `OcrProvider`, and `OcrException`;
- moves crop conversion and preprocessing in front of the provider boundary;
- retains the established Tesseract single-line default, LSTM mode, glyph/layout mapping, and bilinear magnification;
- preserves diagnostic PNG export in the active workspace cache through a provider-neutral writer;
- keeps frame reuse out of recognition settings and models it explicitly at the emulator capture bridge;
- preserves reuse for left-menu multi-region reads and the Bear Trap march-counter read;
- correctly decodes both 32-bit RGBA and 16-bit RGB565 captures;
- removes the unrelated desktop Maven configuration change from the PR.

## Verification

- `mvnw.cmd -pl modules/vision,modules/automation -am test`
- `mvnw.cmd package`
- 428 tests passed, 0 failures, 0 errors
- existing saved-frame OCR tests passed, including deployment cost and both Intel cooldown layouts
- new regression tests cover RGBA channel order, RGB565 conversion, bilinear scaling, diagnostic export, and lazy frame reuse

## Remaining evidence / risk

- No live account-log validation was performed for this maintainer follow-up. Before merging, the updated CI should pass and the affected left-menu/Bear Trap OCR paths should ideally receive a live smoke check.
- This PR does not establish that Paddle is better than Tesseract. That comparison remains Step 2 in #212 and requires ground-truth accuracy and operational measurements on identical prepared inputs.

### DIFF
--- a/modules/api/src/main/java/dev/frostguard/api/domain/OcrSettingsData.java
+++ b/modules/api/src/main/java/dev/frostguard/api/domain/OcrSettingsData.java
@@ -4,67 +4,42 @@ import java.awt.Color;
 import java.util.Objects;
 
 /**
- * Encapsulates the full set of parameters passed to the Tesseract
- * OCR recognition engine. Instances are assembled exclusively
- * through the {@link Configurator} fluent builder.
+ * Encapsulates the full set of parameters passed to the OCR recognition engine.
+ * Instances are assembled exclusively through the {@link Configurator} fluent builder.
  *
  * <p>Pre-configured factories like {@link #forNumberRecognition()}
  * and {@link #forTextBlock()} cover the most common use cases.</p>
  */
-public class TesseractSettingsData {
+public class OcrSettingsData {
 
     /**
-     * Backends supported by the Tesseract recognition engine.
+     * Generic layout hints to inform the OCR engine about the expected
+     * structure of the text to be recognized.
      */
-    public enum RecognitionEngine {
-        LEGACY_ONLY(0), LSTM_ONLY(1), COMBINED(2), AUTO(3);
-
-        private final int code;
-
-        RecognitionEngine(int code) { this.code = code; }
-
-        /** Numeric identifier passed to the native engine. */
-        public int code() { return code; }
-    }
-
-    /**
-     * Page segmentation strategies that control how Tesseract
-     * identifies text regions within an image.
-     */
-    public enum PageAnalysis {
-        OSD_ONLY(0), AUTO_WITH_OSD(1), AUTO_NO_OSD(2), FULLY_AUTO(3),
-        SINGLE_COLUMN(4), VERTICAL_BLOCK(5), UNIFORM_BLOCK(6),
-        SINGLE_LINE(7), SINGLE_WORD(8), CIRCULAR_WORD(9),
-        SINGLE_GLYPH(10), SPARSE(11), SPARSE_WITH_OSD(12), RAW_LINE(13);
-
-        private final int code;
-
-        PageAnalysis(int code) { this.code = code; }
-
-        /** Numeric identifier passed to the native engine. */
-        public int code() { return code; }
+    public enum TextLayout {
+        SINGLE_LINE,
+        SINGLE_WORD,
+        TEXT_BLOCK,
+        SPARSE,
+        AUTO
     }
 
     /* ---- immutable configuration fields ---- */
 
-    private final PageAnalysis pageAnalysis;
-    private final RecognitionEngine recognitionEngine;
+    private final TextLayout textLayout;
     private final boolean isolateForeground;
     private final Color targetColor;
     private final boolean diagnosticMode;
     private final String allowedGlyphs;
-    private final boolean reuseFrame;
 
     /* ---- private: construction via Configurator only ---- */
 
-    private TesseractSettingsData(Configurator cfg) {
-        this.pageAnalysis      = cfg.pageAnalysis;
-        this.recognitionEngine = cfg.recognitionEngine;
+    private OcrSettingsData(Configurator cfg) {
+        this.textLayout        = cfg.textLayout;
         this.isolateForeground = cfg.isolateForeground;
         this.targetColor       = cfg.targetColor;
         this.diagnosticMode    = cfg.diagnosticMode;
         this.allowedGlyphs    = cfg.allowedGlyphs;
-        this.reuseFrame        = cfg.reuseFrame;
     }
 
     /* ---- pre-configured presets ---- */
@@ -73,10 +48,9 @@ public class TesseractSettingsData {
      * Settings optimised for recognising numeric strings
      * (digits, commas, and decimal points only).
      */
-    public static TesseractSettingsData forNumberRecognition() {
+    public static OcrSettingsData forNumberRecognition() {
         return configurator()
-                .pageAnalysis(PageAnalysis.SINGLE_LINE)
-                .recognitionEngine(RecognitionEngine.LSTM_ONLY)
+                .textLayout(TextLayout.SINGLE_LINE)
                 .allowedGlyphs("0123456789,.")
                 .build();
     }
@@ -84,10 +58,9 @@ public class TesseractSettingsData {
     /**
      * Settings suitable for reading a uniform block of mixed text.
      */
-    public static TesseractSettingsData forTextBlock() {
+    public static OcrSettingsData forTextBlock() {
         return configurator()
-                .pageAnalysis(PageAnalysis.UNIFORM_BLOCK)
-                .recognitionEngine(RecognitionEngine.LSTM_ONLY)
+                .textLayout(TextLayout.TEXT_BLOCK)
                 .build();
     }
 
@@ -95,10 +68,9 @@ public class TesseractSettingsData {
      * Settings for recognising isolated single words, such as
      * button labels or short status indicators.
      */
-    public static TesseractSettingsData forSingleWord() {
+    public static OcrSettingsData forSingleWord() {
         return configurator()
-                .pageAnalysis(PageAnalysis.SINGLE_WORD)
-                .recognitionEngine(RecognitionEngine.LSTM_ONLY)
+                .textLayout(TextLayout.SINGLE_WORD)
                 .build();
     }
 
@@ -106,10 +78,9 @@ public class TesseractSettingsData {
      * Settings for reading white text on dark game backgrounds,
      * with foreground isolation enabled.
      */
-    public static TesseractSettingsData forWhiteTextOnDark() {
+    public static OcrSettingsData forWhiteTextOnDark() {
         return configurator()
-                .pageAnalysis(PageAnalysis.SINGLE_LINE)
-                .recognitionEngine(RecognitionEngine.LSTM_ONLY)
+                .textLayout(TextLayout.SINGLE_LINE)
                 .isolateForeground(true)
                 .targetColor(Color.WHITE)
                 .build();
@@ -117,17 +88,15 @@ public class TesseractSettingsData {
 
     /**
      * Returns {@code true} when this settings instance uses all
-     * default (unset) values — no page analysis, no engine, no
+     * default (unset) values — no layout hint, no
      * foreground isolation, and no glyph filter.
      */
     public boolean isDefaultConfiguration() {
-        return pageAnalysis == null
-                && recognitionEngine == null
+        return textLayout == null
                 && !isolateForeground
                 && targetColor == null
                 && !diagnosticMode
-                && (allowedGlyphs == null || allowedGlyphs.isEmpty())
-                && !reuseFrame;
+                && (allowedGlyphs == null || allowedGlyphs.isEmpty());
     }
 
     /**
@@ -138,22 +107,17 @@ public class TesseractSettingsData {
      */
     public Configurator toConfigurator() {
         return new Configurator()
-                .pageAnalysis(this.pageAnalysis)
-                .recognitionEngine(this.recognitionEngine)
+                .textLayout(this.textLayout)
                 .isolateForeground(this.isolateForeground)
                 .targetColor(this.targetColor)
                 .diagnosticMode(this.diagnosticMode)
-                .allowedGlyphs(this.allowedGlyphs)
-                .reuseFrame(this.reuseFrame);
+                .allowedGlyphs(this.allowedGlyphs);
     }
 
     /* ---- primary accessors ---- */
 
-    /** Active page segmentation strategy. */
-    public PageAnalysis pageAnalysis()           { return pageAnalysis; }
-
-    /** Active recognition backend. */
-    public RecognitionEngine recognitionEngine()  { return recognitionEngine; }
+    /** Expected layout of the text. */
+    public TextLayout textLayout()               { return textLayout; }
 
     /** Whether background removal is applied before recognition. */
     public boolean isolateForeground()           { return isolateForeground; }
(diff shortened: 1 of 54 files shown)


modules/vision/src/main/java/dev/frostguard/vision/ocr/OcrEngine.java
```diff
@@ -0,0 +1,111 @@
...
+    private OcrEngine() {}
+
+    private static OcrProvider getProvider() {
+        OcrProvider p = provider;
+        if (p == null) {
+            synchronized (OcrEngine.class) {
+                p = provider;
+                if (p == null) {
+                    log.debug("Initializing Tesseract OCR provider");
+                    p = new TesseractOcrProvider();
```
bizulk, line 40: The provider abstraction is in place, but I don't see the switching mechanism yet. getProvider() always instantiates TesseractOcrProvider, so the implementation is effectively hard-coded to Tesseract. Even if Tesseract is the only supported backend for now, would it make sense to introduce the provider selection/factory hook here so adding another backend later doesn't require changing OcrEngine itself?

bizulk (dismissed):
need to bring clarfication.

modules/vision/src/main/java/dev/frostguard/vision/ocr/OcrEngine.java
Cloudbed-invi: Thanks for the review! I intentionally left the switching mechanism (like an OcrEngineType enum and setActiveEngine() setter) out of this PR to keep Step 1 strictly as a pure, zero-behavior-change interface extraction.

The plan was to introduce the switching hook in Step 2 alongside the actual RapidOCR/Paddle implementation, so that we aren't merging dead code / unused setters into main right now.

However, if you'd prefer to have the OcrEngineType enum and the factory selection hook added to OcrEngine in this PR to finalize the API's shape early, I can easily push that change! Just let me know if you want that included here, or if it's okay to wait for Step 2.

bizulk: Hello.
I would prefer, because otherwise you would be compelled to make a lot of change again to include this selection.
Also that would help me to review what you foresee for the engine lifecycle.
Thank you.

CodeLtDave:
Thanks for splitting #135 into smaller steps. This PR now matches the intended scope much better.

Regarding the open discussion: I would keep runtime provider selection out of Step 1. With only one provider, an `OcrEngineType` enum or `setActiveEngine()` API would be unused and its lifecycle behavior could not be meaningfully tested yet. Step 2 should introduce provider selection together with the actual second provider, its fallback behavior, and lifecycle tests.

Before merging Step 1, however, I would like to make the abstraction genuinely provider-neutral:

- `OcrProvider` should not expose `TesseractSettingsData`. Either introduce backend-neutral OCR options, or perform the shared preprocessing before delegating to a provider.
- `OcrEngine` should not expose or wrap failures as `TesseractException`; the public facade needs a backend-neutral error contract.
- `toBufferedImage()` is image conversion rather than OCR behavior and should be moved to a shared conversion utility instead of being implemented by every OCR provider.
- The interface currently requires every implementation to be thread-safe and describes one shared instance per provider type. I would avoid fixing that lifecycle contract before Step 2 demonstrates what the RapidOCR backend actually supports.

The existing call-site migration and the preservation of Tesseract behavior look appropriate for Step 1. Once the remaining Tesseract-specific parts are removed from the public abstraction and CI remains green, I would consider this step mergeable without adding the runtime switch yet.

CodeLtDave (changes requested):
finish abstraction as outlined in my comment

bizulk:
What is the point of this step1 : 
- clarify the contrat the interface expose : agreed. Tipically you brought the right comments. 
- clarify the engine lifecycle : couldn't we chek this with one engine only ?

About the change pushing to step 2 the engine selection : I was wrong, this would only impact at creation, not the classes consuming the service.

... Finally I'm okay with this, I don't believe anymore step2 would break the API.

Cloudbed-invi:
I've updated this PR to focus exclusively on Step 1. I removed the runtime provider selection for now and focused on making the OcrProvider abstraction genuinely provider-neutral (renamed TesseractSettingsData to OcrSettingsData, replaced TesseractException with OcrException, and moved toBufferedImage to a utility class). I will follow up with Step 2 (adding the new provider and selection mechanism) once this is merged

CodeLtDave:
Thanks for the quick update. Moving image conversion out of the provider, introducing `OcrException`, and keeping runtime selection in Step 2 are all improvements in the right direction.

I still see one architectural blocker before I can approve this step: `OcrSettingsData` is currently mostly a rename of `TesseractSettingsData`, rather than a provider-neutral contract.

It still exposes Tesseract-specific concepts and native values:

- `RecognitionEngine` contains `LEGACY_ONLY`, `LSTM_ONLY`, and `COMBINED`.
- `PageAnalysis` exposes numeric Tesseract page-segmentation codes.
- Accessors such as `getPageSegMode()` and `getOcrEngineMode()` remain part of the shared API.
- The class documentation still describes parameters passed to Tesseract.
- `reuseFrame` concerns screenshot/capture ownership rather than OCR provider behavior.
- Shared cropping, scaling, and foreground isolation still live inside `TesseractOcrProvider`.

With this contract, the Paddle provider would still need to interpret Tesseract settings, silently ignore them, or duplicate Tesseract's preprocessing. That is the same compatibility risk identified in #135.

I suggest defining the provider boundary around a prepared image and genuinely generic recognition hints. For example, `OcrEngine` or a shared preprocessing component can handle cropping, scaling, foreground isolation, and diagnostic preparation before delegating a `BufferedImage` to the provider. Generic hints may include language, allowed characters, and semantic layouts such as single line, single word, or text block. Native PSM codes and LSTM/legacy selection should remain internal to `TesseractOcrProvider`.

The lifecycle point also remains partially unresolved: `OcrEngine` still caches one provider for the JVM lifetime and states that all methods are thread-safe. I would avoid making that promise until Step 2 can validate what the RapidOCR provider actually supports.

There are also a few minor cleanup items:

- `OcrException` currently says it wraps provider-specific exceptions such as `OcrException`.
- `OcrEngine` contains redundant catch-and-rethrow blocks and an unused import.
- The PR description is now outdated and still lists `toBufferedImage()` as part of `OcrProvider`.

Once the public settings contract and preprocessing boundary are genuinely provider-neutral, the lifecycle claim is corrected, and CI plus the existing saved-frame tests remain green, I would consider Step 1 ready to merge.

modules/api/src/main/java/dev/frostguard/api/domain/OcrSettingsData.java
bizulk, line 1: It is hard coded configuration. Is this sufficient to rename it to a generic name ? 
What happens if we want another variant  ? 
Tipically will the paddle reuse it use or something else, that would make this naming unsuitable ?
Is there also a step2 for the configuration selection ?
I need to clarify this before approving. Thank you.

bizulk (dismissed):

modules/api/src/main/java/dev/frostguard/api/domain/OcrSettingsData.java
Cloudbed-invi: That is a very good catch. You're completely right that the internal enums (like PageAnalysis mapping to Tesseract PSMs, and RecognitionEngine mapping to OEMs) still heavily reflect Tesseract's specific configuration.

My plan for Step 2 is to treat OcrSettingsData as a high-level, intent-based configuration. For example:

Layout Intent: A flag like PageAnalysis.SINGLE_LINE maps to Tesseract PSM 7. But for PaddleOCR, the new provider can interpret this same intent by skipping the text-detection phase entirely and directly running the recognizer on the cropped image (since we already know it's a single line).
Pre-processing: Flags like isolateForeground and targetColor are already engine-agnostic and will apply to the image buffer before it's handed to PaddleOCR.
Whitelisting: For allowedGlyphs, if PaddleOCR doesn't natively support dynamic per-request whitelists as easily as Tesseract, the Paddle provider implementation can interpret this intent by applying a regex/filter to the output text.
By renaming it to OcrSettingsData in this Step 1 PR, we secure a provider-neutral method signature across the ~35 call sites in our tasks. In Step 2 (which will introduce the Paddle provider and the configuration selection), we will evolve the internals of OcrSettingsData to use truly generic naming (e.g., ExpectedLayout instead of PageAnalysis) and map those intents inside each provider.

Does this phased approach sound good to you for Step 1? Or would you prefer I also sanitize the internal fields and enums of OcrSettingsData right now in this PR before we merge?

bizulk: its just that for the engine you insterted a generic wrapper, here it is just renaming. I would have expected you use a generic mechanisme to forward to the right implementation.
if you believe bringe paddle configuration won't make you replay this PR that is ok.
Let's see how things go.

Cloudbed-invi: Thank you for the feedback!

To answer your questions: yes, there is absolutely a Step 2 planned for the configuration selection. The reason the configuration wrapper currently just acts as a renaming (and still has some hard-coded Tesseract fields) is because another maintainer previously requested that I keep runtime provider selection strictly out of Step 1. Their reasoning was that with only one provider currently implemented, any selection or fallback API (like an OcrEngineType enum) would be unused and its lifecycle behavior couldn't be meaningfully tested yet.

In Step 2, I will be introducing the Paddle/RapidOCR backend. Along with that, I'll add the actual provider selection mechanism, fallback behavior, and fully sanitize OcrSettingsData so it dynamically forwards to the correct implementation's configuration. This shouldn't require replaying the current PR, but rather building directly on top of the generic OcrProvider interface we've established here.

Also, as a heads up: I just pulled the latest main and resolved the merge conflicts caused by the recent directory restructuring, so the branch is fully up to date and all tests are passing again. Let me know if you are good with this approach for Step 1!

CodeLtDave (changes requested):
Thanks for updating the branch and resolving the module-layout conflicts from the earlier `main`. I think there is still a misunderstanding in how my previous feedback divides Step 1 from Step 2, so I want to make the required boundary explicit.

Runtime provider selection belongs in Step 2. I am not asking for an `OcrEngineType`, a Paddle provider, fallback behavior, or UI/configuration selection in this PR.

The provider-neutral contract itself belongs in Step 1, however. That is the purpose of this PR, and it should not be deferred to Step 2. At the moment, `OcrSettingsData` remains a renamed Tesseract configuration:

- `RecognitionEngine` exposes Tesseract's legacy/LSTM/combined modes.
- `PageAnalysis` exposes Tesseract PSM values as native integer codes.
- `getPageSegMode()`, `getOcrEngineMode()`, and the compatibility aliases keep those details in the shared API.
- `reuseFrame` belongs to capture ownership in the automation layer, not to an OCR provider request.
- Cropping, scaling, foreground isolation, and diagnostic preparation still happen inside `TesseractOcrProvider`.

Deferring this cleanup means Step 2 must redesign the public settings contract and preprocessing boundary immediately after Step 1 merges. It also leaves Paddle with exactly the unsafe choices identified in #135: interpret Tesseract options, ignore them, or duplicate Tesseract preprocessing.

Please complete the abstraction in this PR:

1. Define generic, semantic OCR input/options, such as language, allowed characters, and expected layout (`SINGLE_LINE`, `SINGLE_WORD`, `TEXT_BLOCK`) without native Tesseract codes or engine modes.
2. Move provider-independent crop/scaling/foreground preprocessing before the provider boundary and pass a prepared image/input to the provider.
3. Keep Tesseract PSM/OEM mapping internal to `TesseractOcrProvider`. Any genuinely provider-specific tuning must not leak through the shared contract.
4. Move `reuseFrame` handling to the capture/automation layer.
5. Remove the unverified global thread-safety promise. The current Tesseract singleton may remain an implementation detail for Step 1, but the abstraction should not prescribe lifecycle guarantees for future providers before they are tested.
6. Clean up the incorrect `OcrException` documentation, redundant catch-and-rethrow blocks, and unused imports.

After that, please update the PR description to match the actual API, bring the branch onto current `main`, resolve the current conflicts, and rerun CI plus the existing saved-frame OCR tests. The current green checks are from August 12 and predate the current `main`.

This multi-step architectural work should also be linked to a planning issue/project item that records the three-step scope, acceptance criteria, dependencies, and evidence requirements.

Once Step 1 provides that stable neutral boundary, Step 2 can add Paddle, provider selection, lifecycle/fallback behavior, native packaging, and provider-specific verification without replaying this PR.

Cloudbed-invi:
I've pushed updates to fully address the feedback for Step 1. The abstraction is now genuinely provider-neutral:

1. **Provider-Neutral Settings:** `OcrSettingsData` (formerly `TesseractSettingsData`) no longer exposes Tesseract-specific enum codes (`RecognitionEngine`, `PageAnalysis`) or engine modes. It now uses generic semantics like `TextLayout` (`SINGLE_LINE`, `TEXT_BLOCK`, `SINGLE_WORD`, etc.), `isolateForeground`, and `targetColor`.
2. **Preprocessing Boundary Moved:** Provider-independent image preprocessing (cropping, scaling with `NEAREST_NEIGHBOR` to preserve font sharpness, and foreground isolation) has been moved out of the OCR provider and into `OcrEngine`. The provider now receives a fully prepared image.
3. **Internal Tesseract Mapping:** Tesseract-specific PSM/OEM mappings and initialization logic are now kept strictly internal to `TesseractOcrProvider`.
4. **Separated Image Conversion:** `toBufferedImage()` was moved to a separate `ImageConverter` utility class, and `reuseFrame` handling was removed from the OCR provider boundary.
5. **Exception Handling:** Cleaned up `OcrException` documentation and removed redundant catch-and-rethrow blocks.
6. **Testing & Regressions:** I've resolved conflicts with `main`, restored Tesseract's default `AUTO` PSM, and verified that all existing saved-frame OCR tests (`PetSkillsCooldownOcrFrameTest`, `DeploymentCostOcrFrameTest`, etc.) pass successfully with these changes.

The PR is now strictly focused on establishing a clean interface. Runtime provider selection and the new Paddle/RapidOCR implementation will follow in Step 2. 

Let me know if this looks good to merge!

CodeLtDave:
I pushed the agreed maintainer follow-up as eb47e4b9. It restores the behavior that was still lost across the abstraction (diagnostic output, bilinear preprocessing, the single-line default, and explicit frame reuse), fixes RGBA/RGB565 conversion, removes the unrelated desktop POM change, and adds focused regression coverage. The full reactor build passes locally: 428 tests, 0 failures/errors. I have not done live account-log validation, so I would still smoke-test the left-menu queue reads and Bear Trap march counter before merge. Part of #212.

CodeLtDave (approved):
The requested Step 1 boundary is now implemented, the regression fixes are covered, and both Linux and Windows installer CI pass on eb47e4b9. My earlier change request is addressed. Live smoke validation of the left-menu queue reads and Bear Trap march counter is still recommended before merge.

CodeLtDave (approved):
Reviewed the rebased single-commit diff on current main. The Arena power-ranking changes from #204 are preserved through the neutral OCR API. Local focused tests and full package pass (431 tests), and the replacement SHA passes Linux plus Stable/Nightly Windows installer CI. Live account-log validation remains unperformed and documented.

CodeLtDave:
Rebased onto current main as 24768a4c and consolidated the PR into one coherent commit while preserving Cloudbed_invi as co-author. The only semantic conflict was the newer Arena power reader from #204; its ranking and crop behavior remain intact and its OCR call now uses the neutral boundary. Verification on the rebased SHA: focused vision/automation/tasks tests, full mvnw.cmd package (431 tests), Linux CI, and Stable/Nightly Windows packaging + launcher/workspace smoke tests all pass. Live account-log validation was not performed.

PR #158, opened 2026-08-10
type(scope): implement profile and settings export/import with safe merge

Cloudbed-invi:
### type(scope): implement profile and settings export/import with safe merge

**Summary**
This PR implements full configuration export, import, and reset functionality for user profiles and global settings in the `ProfileManagerActionController` and `ProfileTransferService`.

**Changes:**
- Added robust Profile Export/Import using Jackson JSON serialization, packaging both `AccountDescriptor` profiles and active global settings.
- Implemented a "Safe Export" mode that automatically strips sensitive `TELEGRAM_` configuration keys from the payload to prevent accidental leakage of Bot Tokens and Chat IDs.
- Introduced ID-based merging on import: instead of generating duplicates, importing a profile that already exists locally will merge and override its settings gracefully.
- Added a "Reset Config" action to selectively wipe configurations of an existing profile back to default states without removing the profile entry itself.
- Updated the Import Success UI alert to prompt users to restart their application so that imported global settings correctly populate in the `LauncherLayoutController` panes.

**Evidence:**
- Automated tests pass on reactor build (`mvn package`).
- Verified live account-log confirmation for successful database persistence of imported settings and correct exclusion of Telegram tokens during "Safe Export".

### DIFF
--- a/fg-app/src/main/java/dev/frostguard/app/panel/profile/ProfileManagerActionController.java
+++ b/fg-app/src/main/java/dev/frostguard/app/panel/profile/ProfileManagerActionController.java
@@ -151,7 +151,7 @@ public class ProfileManagerActionController implements ProfileStatusChangeListen
 						}
 					}
 				} catch (Exception ex) {
-					// Ignore invalid global configs
+					log(TpMessageSeverityEnum.WARNING, "Failed to apply global configuration " + gCfg.name() + ": " + ex.getMessage());
 				}
 			}
 		}
--- a/fg-app/src/main/java/dev/frostguard/app/panel/profile/ProfileManagerActionController.java
+++ b/fg-app/src/main/java/dev/frostguard/app/panel/profile/ProfileManagerActionController.java
@@ -119,9 +119,17 @@ public class ProfileManagerActionController implements ProfileStatusChangeListen
 					globalConfigs.addAll(transfer.globalConfigs());
 				}
 				for (AccountDescriptor descriptor : transfer.profiles()) {
-					String name = availableName(descriptor.getName(), occupiedNames);
-					descriptor.setName(name);
-					occupiedNames.add(name);
+					java.util.Optional<ProfileAux> existing = existingProfiles.stream()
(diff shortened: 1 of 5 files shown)


CodeLtDave (changes requested):
Thanks for the contribution. The core profile import/export functionality is already present on `main` from `d19e764` (including profile configs, tags, validation, multi-profile transfer, preview, and collision-safe naming), so this PR currently duplicates a substantial part of the existing implementation.

Please rework this as a focused change on top of the current `main` and address the following before it can be merged:

1. **Clarify and make the merge behavior safe.** The implementation is described as ID-based, but exported IDs are intentionally omitted. It actually matches profiles by case-insensitive name and then assigns the local ID, causing the complete local profile, configs, and tags to be overwritten. This must not happen implicitly. The import preview should make the choice explicit (for example: create, replace, or skip), or use a real stable identity with clearly defined semantics.
2. **Do not leak secrets in ?Safe Export?.** Filtering only keys beginning with `TELEGRAM_` still exports other credentials such as `DISCORD_TOKEN_STRING`. Please use an explicit sensitivity policy/allowlist rather than a Telegram-specific prefix check.
3. **Separate portable settings from machine-local state.** Importing all global settings currently overwrites values such as emulator paths, the current emulator, and autostart/runtime settings. Define which global settings are portable and let users review or select them before applying the import.
4. **Handle reset failures accurately.** `resetProfileConfigs` ignores the result of `saveProfile`, while the UI always reports success. Failed or partial resets must be surfaced and should not be reported as successful.
5. **Add targeted tests for the new behavior.** In particular, cover safe export (including non-Telegram secrets), global-setting round trips and legacy files, new/existing/same-name profile imports, explicit overwrite behavior, and reset persistence/failure handling. The current test change only adapts the existing round-trip test to the new return type and does not verify these additions.
6. **Rebase onto current `main`.** GitHub currently reports this PR as conflicting, and the project layout has since moved to `modules/*`.

The global-backup and reset ideas can be useful, but they need to be separated from the already implemented profile transfer and given explicit safety semantics before merge.

PR #249, opened 2026-08-18
Apply enabled-task checkbox changes live, no restart required

trophyscar-bit:
Toggling a task's enabled checkbox only ever took effect on the next app launch -- `ScheduleService.prepareQueue()` reads each task's enabled-bool config exactly once, at engine start, to build the fixed TaskQueue. Flipping a checkbox mid-session persisted the new value to the DB fine, but nothing told the already-running queue a task had just become eligible.

Custom tasks already solved this exact problem for their own registry (`CustomTasksLayoutController.scheduleTask/unscheduleTask` against the live `TaskDispatcher`). This generalizes the same live-apply pattern to every standard config-bool-gated daily task:

- `ScheduleService.applyEnabledTaskChange(AccountDescriptor, ConfigurationKeyEnum, boolean)`: finds the `TpDailyTaskEnum`(s) whose `getConfigKey()` matches, and either enqueues them fresh or evicts them against the account's live queue. No-ops safely (logged, not an error) if the bot isn't running for that profile.
- `ProfileManagerLayoutController`: after a boolean config save succeeds, fetches the fresh `AccountDescriptor` and calls `applyEnabledTaskChange`. No new UI needed -- every settings panel already auto-saves per-toggle, so this is a single choke point for every checkbox in the app.

No shared-infrastructure dependency -- compiles and applies standalone. Compiled + full test suite green on my fork before opening this.

### DIFF
--- a/modules/automation/src/main/java/dev/frostguard/engine/service/ScheduleService.java
+++ b/modules/automation/src/main/java/dev/frostguard/engine/service/ScheduleService.java
@@ -383,6 +383,65 @@ public class ScheduleService {
 		}
 	}
 
+	/**
+	 * matt/2026-08-16: "if I click a box, like, that task won't come up... it doesn't just
+	 * automatically enable it" -- {@link #prepareQueue} only ever reads each task's enabled-bool
+	 * config ONCE, at engine launch, so toggling a checkbox mid-session persisted the new value
+	 * to the DB (via AbstractProfileController -&gt; notifyProfileChange -&gt; saveProfile) but never
+	 * touched the already-running {@link TaskQueue} -- a restart was genuinely the only thing that
+	 * picked it up. Custom tasks already solve this exact problem live (see
+	 * CustomTasksLayoutController.scheduleTask/unscheduleTask calling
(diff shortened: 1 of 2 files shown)


CodeLtDave:
Thanks for the contribution. The reported issue is valid, but it was already fixed centrally in #186.

Profile setting writes now pass through `ConfigService.writeAccountSetting(...)`, which reloads the committed account and calls `ScheduleService.applyCommittedProfileSetting(...)`. That path already updates the running profile and reconciles config-driven tasks, including changes made outside the profile UI.

This PR would apply UI changes a second time. It would also bypass existing safeguards—for example, `SKIP_TUTORIAL_ENABLED_BOOL` is intentionally restart-only, while this implementation would enqueue `SKIP_TUTORIAL` in the active session. It also does not check whether a task is currently executing before enqueueing another instance.

Because the requested behavior is already present and the additional implementation should not be merged, I’m closing this as superseded by #186.

I verified the branch with `mvnw.cmd -pl modules/desktop -am test`: 510 tests passed with no failures.

PR #251, opened 2026-08-18
Statistics: add a full earnings/activity dashboard with a timeframe dropdown

trophyscar-bit:
**Tracking:** Part of #279 (workspace-scoped telemetry and statistics reporting). Depends on #250, which must merge first.

## Summary

A full earnings/activity dashboard for the Statistics tab, driven by a timeframe dropdown -- Past
hour, Last night, Last 24 hours, This week, This month, All time -- reading `TelemetryReport`.

### Scope correction (2026-09-02)

An earlier version of this description understated the change as a UI-only addition. That was wrong:
this PR makes substantial `TelemetryReport` changes beyond #250, and the real diff is **35 files**
(it also carries #250's commit, since it stacks on it). Past Hour and This Month are new windows;
This Week is the same rolling-7-day maths the design already used, relabelled for the dropdown.

## Zero change versus not enough data

These are now decided separately, which is the correction from the last round:

- `computeViewState` decides "nothing changed" from whether **any earned delta or activity count is
  non-zero**, not from whether the lists are empty. Emptiness stopped meaning "no change" once an
  unchanged metric began emitting a zero-valued `Delta`; the earned list is then non-empty, so the
  zero-change branch was unreachable on real data while real unchanged windows fell through to the
  normal branch.
- Coverage answers the separate question of whether the window could be measured at all, which is
  what distinguishes "nothing happened" from "nothing is known yet". Activity carries its own
  sufficiency information rather than inferring it from the presence of positive changes.
- `StatisticsLayoutWiringTest` parses the FXML and reflects over the controller (every `fx:id` has a
  field, every `onAction` has its method, refresh and the timeframe selector are present).
  Deliberately structural: a live scene-graph test needs Monocle or Xvfb on headless CI.

```
upstream/main

upstream/main

upstream/main
```

```
```

The Statistics tab exists (`StatisticsLayout.fxml`, `StatisticsLayoutController`) and shows
current values. There is no earnings or activity history and no way to ask for a time window,
because there is no stored history to query.

### What this PR does

Adds an earnings/activity dashboard driven by a timeframe dropdown -- Past hour, Last night,
Last 24 hours, This week, This month, All time -- reading `TelemetryReport` from #250.

### The difference

Upstream answers "what is my furnace level now". This answers "what did the bot actually earn
and do between these two times".

It makes the bot auditable. Without a window you cannot tell a productive night from a night
where a routine silently stalled. Past Hour and This Month are new; This Week is the same
rolling-7-day maths already in the design, relabelled for the dropdown. All windows reuse
`TelemetryReport`'s generic `last()` / `activityLast()` / `coverageForLast(amount, unit)`, so no
report-engine changes ride along with the UI.

### Verification status

DCO green. Java CI has never run on this branch. **Depends on #250** -- merge that first or
land them together; this branch stacks on #250's commit `0683533`, which is why its diff carries
the telemetry files too.

## Validation on this head (`e11b148`)

Measured, with the command and counting method stated so the number is checkable:

```
mvn -o test          # Temurin JDK 21, full reactor, no packaging
```

**698 tests, 0 failures, 0 errors, 0 skipped**, across 188 suites. Counted by summing the
`tests`/`failures`/`errors`/`skipped` attributes of every
`**/target/surefire-reports/TEST-*.xml`, not read off a single module's summary line.

The two classes covering the points raised in review:

| Class | Tests |
|---|---|
| `StatisticsLayoutControllerTest` | 13 |
| `StatisticsLayoutWiringTest` | 5 |


CodeLtDave (changes requested):
Thanks for the contribution. The dashboard direction is useful, but there are several merge-blocking issues that need to be addressed before this can land.

1. **Use the selected workspace and a stable profile identity.** \`TelemetryReport.load()\` currently reads \`telemetry/history.jsonl\` below \`System.getProperty(user.dir)\`, even though runtime data must live below \`WorkspacePaths.current()\`. It also filters by mutable/non-unique profile name, and rows without a profile are accepted for every profile. Please store and load telemetry by stable profile ID within the selected workspace so Stable/Nightly and concurrent accounts remain isolated.

2. **Correct the timeframe boundary math.** \`deltaOverWindow()\` chooses the first sample at or after the cutoff as its baseline. With #250's hourly writer, “Past hour” will normally contain only the most recent sample, making start and end identical and producing no delta. Please use explicit, tested boundary semantics—normally the last valid metric value at or before the window start and the last valid metric value at or before the end.

3. **Make coverage and metric endpoints truthful.** The report uses one global end sample even when individual metrics are missing, while the UI reads current values only from the newest global sample. A transient missing metric can therefore disappear despite valid history, and the displayed coverage can describe timestamps different from those used for a metric. Unchanged values are also collapsed into the same “not enough samples” state as genuinely insufficient history. Please model metric-specific endpoints/coverage and distinguish zero change from missing evidence.

4. **Resolve the dependency and scope before merge.** This PR has no telemetry writer and is empty in normal use if merged alone. Its \`TelemetryReport.java\` is byte-identical to #250, which currently has requested changes for the same storage/profile/correctness problems, and #250 also overlaps #254's stockpile/speedup keys. Please establish a clean dependency stack or combine the native telemetry foundation with its tested consumer. Also update the title/body: current \`main\` has no four-button telemetry row to replace; this PR actually introduces the complete dashboard (865 changed lines), external icon loading, and collapses the existing task statistics.

5. **Add focused validation.** There are no tests in this PR. Please cover workspace resolution, profile isolation, malformed/truncated JSONL, metric-specific missing values, zero-change windows, Past Hour/24-hour/month boundaries, time zones/DST, counter resets, dropdown refresh/profile switching, and FXML/controller behavior. The required #250 OCR writer also needs saved 720×1280 frame coverage and live account-log confirmation. Live validation should include multiple profiles, installed Stable/Nightly workspace paths, an hourly overnight sequence, reset behavior, and the rendered JavaFX layout.

Please also add the missing \`.stat-tile-glyph\` CSS sizing rule (the code relies on it for the 72 px icons), remove the author/date changelog comments in favor of concise rationale, use observable logging instead of silently swallowing all load/parse failures, link the substantial work to an issue/project item, and sign off the commit for DCO.

Local verification against current \`main\` (\`7d2a7549\`) completed successfully: \`mvnw.cmd package\` passed 512 tests with 0 failures/errors/skips and produced the desktop bundle. \`git diff --check\` also passed. These results establish compilation and existing-regression coverage only; the new telemetry/report behavior remains untested.

trophyscar-bit:
Pushed 2 more commits addressing the rest of your review:

- Rebased onto #250's latest (commit 20e4056), which fixes the 'Past Hour always reads zero' bug you flagged -- the baseline now anchors to the last known value before the window opened instead of the first in-window sample.
- Added the missing `.stat-tile-glyph` CSS rule for the 72px icons.
- Added tests (StatisticsLayoutControllerTest + 2 new TelemetryReportTest cases covering the Past-Hour fix specifically).
- Fixed the PR title -- you're right, main has no 4-button row to 'replace'; this adds a new dashboard section.

Full suite: 522 tests, 0 failures.

CodeLtDave (changes requested):
Thank you for the thoughtful follow-up and for addressing a substantial part of the previous review. The workspace/profile-ID isolation, atomic profile-scoped storage, per-metric end selection, resource Past-Hour baseline, 08:30 wake alignment, icon CSS, corrected title, and initial tests are all meaningful improvements.

I reverified the updated head \`86595c9\` against current \`main\`: \`mvnw.cmd package\` passes 522 tests with 0 failures/errors/skips and produces the desktop bundle, \`git diff --check\` passes, and \`examples/custom-tasks/bg_telemetry.java\` compiles separately with Java 21. Thank you for adding that coverage and for documenting what changed.

A few merge blockers remain:

1. **Handle a sample exactly at the window start correctly.** The updated resource baseline is documented as “at or before” the boundary, but the implementation only accepts samples strictly before \`from\` (\`isBefore(from)\`). If an hourly or 23:00 anchor exists exactly at the window start, the report skips it and uses an older sample, overcounting activity before the requested window. Please use the latest metric value at or before the boundary and add an exact-boundary regression test.

2. **Apply the boundary fix to activity deltas as well.** \`activityOverWindow()\` still uses the earliest activity snapshot at or after \`from\`. With the hourly writer, “Past hour” will normally have one activity-bearing sample, so start and end are identical and the Accomplished section remains empty. Activity needs the same tested pre-window/exact-boundary semantics as resource metrics, including reset/decreasing-counter behavior.

3. **Make displayed coverage and card values match the samples actually used.** Resource deltas now use metric-specific endpoints, but \`coverageForWindow()\` still reports global first/last samples, and the controller obtains every current card value only from \`telemetry.latest()\`. A metric missing from the newest global row can therefore disappear despite a valid metric-specific end value, while the displayed coverage describes different timestamps. The UI also still reports “not enough samples” when there is sufficient history but all changes are zero. Please model sufficient-data/zero-change separately and make coverage/current values derive from the relevant metric endpoints.

4. **Complete end-to-end evidence for the required writer.** This branch now contains #250, but the OCR/navigation writer still lacks saved 720×1280 frame coverage and live account-log confirmation. The fixed taps, HUD crops, plausibility/retry behavior, multi-profile writes, an hourly overnight sequence, installed Stable/Nightly workspace paths, reset behavior, and the rendered JavaFX dashboard remain unverified. The dashboard should not be considered merge-ready until its data source is demonstrated end to end. Please also clarify the installation/enabling path for the example custom task and its dependency on the stockpile values from #254.

5. **Finish targeted tests and repository gates.** The new controller test only exercises coverage-string formatting; it does not load the FXML or test dropdown selection, refresh, profile switching, missing metrics, zero-change state, or rendered cards. The PR still has no issue/Project item, DCO still fails because the original commits lack matching \`Signed-off-by\` trailers, and Java/Windows CI has not run for the current #251 head. The PR title is corrected, but the body still says it replaces a four-button row that does not exist on \`main\`. Author/date changelog comments also remain, and the drop-in icon path still resolves from \`user.dir\` rather than the selected workspace.

This is good progress and the feature direction is still welcome. I am keeping Changes requested only for the remaining correctness, end-to-end evidence, and repository-gate items above.

trophyscar-bit:
Rebased directly onto #250's squashed, signed tip -- this diff is now genuinely isolated to what #251 owns (StatisticsLayoutController.java, StatisticsLayout.fxml, style.css + tests), not a byte-identical copy of TelemetryReport.java/bg_telemetry.java. GitHub will still show #250's content as part of this PR's diff until #250 merges (cross-fork PRs can't target another fork's branch as base) -- once it does, this PR's diff collapses to just its own ~670 lines.

Two of the round-2 blockers came through the #250 rebase for free (the exact-window-boundary fix and the removed dead comments). Fixed the two that are genuinely this PR's own:

1. **Coverage/current-value alignment**: added `TelemetryReport#latestValueOf(metric)` -- each resource card's "current" value now finds that metric's own most recent sample independently instead of reading the single latest overall sample. A metric missing from just the newest row no longer disappears from its card.
2. **"No data" vs "zero change"**: `showWindow()` now checks whether `coverage` is non-null to tell "genuinely insufficient telemetry" apart from "the window was covered but nothing actually changed" -- these used to render as the identical "not enough samples" message.

Also confirmed `.stat-tile-glyph` CSS is present (was flagged missing in an earlier round) and corrected the PR body -- there's no existing 4-button row on `main` to "replace", this adds a new dashboard.

Still outstanding: end-to-end evidence for the underlying OCR/navigation writer is #250's scope, not this PR's. No FXML-loaded controller test (dropdown selection, refresh, profile switching, rendered cards) -- that needs a JavaFX-toolkit test harness this module doesn't have set up, flagging rather than skipping silently.

Rebuilt clean, full suite (543 tests) green.

CodeLtDave (changes requested):
Thank you for the substantial follow-up work. The rebase onto #250's signed squash commit, the corrected window boundary and activity baseline, the metric-specific latest-value handling, the cleaned-up PR description, and the tests are all meaningful improvements. I also confirmed that the current Linux, Windows installer, and DCO checks are green.

I am keeping **Changes requested** for the following remaining items:

1. **Rolling windows still misclassify a valid zero-change interval as insufficient data.**
   `coverageForWindow(from, to)` derives coverage from `deltaOverWindow(...)` and returns `null` when the delta lists are empty. The controller then treats `bothEmpty && coverage == null` as insufficient data. As a result, a window with usable start/end samples but unchanged values still renders "not enough samples" instead of "no measurable change". The existing `coverageForWindowIsNullWhenThereAreNoDeltas` test currently codifies this behavior. Please derive sufficiency/coverage from usable sample endpoints independently of whether the numeric delta is zero, and add a regression test for an adequately sampled unchanged rolling window through the controller/view-state contract.

2. **All-time coverage is still global rather than metric-specific.**
   `coverageForTotal()` uses the report's global first and last samples, so the displayed range can begin before a metric was first captured or extend beyond its last valid sample. Please make the all-time coverage reflect the actual resource/activity evidence used for the displayed totals and cover this with tests.

3. **Activity counters need coverage for missing keys and resets.**
   Activity calculation still selects one global activity-bearing start/end sample. A counter that first appears inside the window can therefore be omitted, and decreasing/reset counters are silently dropped. Please define and test the intended conservative behavior for newly appearing, missing, and reset counters so that activity is not accidentally underreported or presented with misleading coverage.

The automated build is healthy locally as well (`mvnw.cmd package`: 533 tests, 0 failures/errors/skips; `examples/custom-tasks/bg_telemetry.java` also compiles separately with Java 21). The remaining automated gap is a real FXML-loaded Statistics controller test covering timeframe selection, refresh/profile changes, and the zero-change versus insufficient-data presentation; the current helper-focused controller tests and static FXML binding check do not exercise that runtime path.

Finally, the saved-frame and live evidence called out by the repository rules is still missing: representative 720x1280 HUD/OCR frames, navigation/tap assumptions, OCR plausibility/retry behavior, hourly/overnight collection, reset behavior, multiple profiles/workspaces, and confirmation of the rendered dashboard from real account logs. Since #251 depends on #250's writer, that dependency should be approved/merged and this evidence recorded before #251 is considered merge-ready.

Once the zero-change/coverage semantics and their regression tests are fixed, the remaining work should be mostly validation and dependency closure rather than another broad rewrite.

trophyscar-bit:
Rebased onto #250's fixed tip (e2181c2) — item 4 (zero-change/no-data coverage) is inherited for free, same fix applied.

Addressed items 5 and 6 from the round-3 review:

**5. coverageForTotal() truthfulness — fixed.** Used to return the report's global first/last sample unconditionally, even for a metric that wasn't actually captured at either boundary. Now derives from `coverageForWindow()` over the same bounds `total()` itself uses, so the range is the real per-metric envelope instead of a hardcoded pair. New test confirms it still resolves to the true global bounds when a metric genuinely spans the whole report, and a separate test confirms null with fewer than 2 samples.

**6. Activity counter underreporting — fixed.** `activityOverWindow()` picked one shared start/end sample pair for every counter, which caused two real bugs: a counter missing from just the *chosen* sample (added to tracking later, or a transient write gap) got skipped even though other samples had it, and a counter that reset mid-window produced a negative diff the old `change > 0` filter silently dropped. Rewrote to match `deltaOverWindow`'s per-metric approach — each counter finds its own baseline/end independently; a reset (negative diff) now reports the real post-reset activity (the end value) instead of silence or a misleading negative number. 4 new tests cover: counter missing from the chosen baseline sample but found elsewhere, a genuinely new counter falling back correctly, a counter never captured staying correctly absent, and the reset case.

**Item 7** (a real FXML-loaded controller test exercising timeframe selection/refresh/zero-change-vs-insufficient-data) genuinely needs a JavaFX-toolkit test harness this module doesn't have set up — flagging as real, not attempting a rushed version tonight. **Item 8**'s evidence remains #250's scope.

25/25 `TelemetryReportTest` cases pass (19 pre-existing + 6 new).

trophyscar-bit:
**Round-3 items 7 & 8**

**Item 7** — "a real FXML-loaded Statistics controller test (timeframe selection, refresh, zero-change vs insufficient-data) is still missing — genuinely needs a JavaFX-toolkit test harness this module doesn't have set up."

That's accurate for a literal FXML/scene-graph test (TestFX or Monocle) — this module has never had headless-JavaFX test infra, and adding it now for one round of coverage is a real CI-fragility risk I didn't want to take on blind.

Instead, in `a3993b4`: extracted `showWindow`'s entire decision — what `insufficientData`/`zeroChange`/the metric cards/the activity hint should be, given a window's raw data — into a new pure static method, `computeViewState()`, with zero JavaFX types in it. `showWindow` is now a 2-line wrapper: compute the `ViewState`, then paint it onto FXML nodes via `applyViewState()`. Pure extraction, behavior unchanged.

`StatisticsLayoutControllerTest` now drives `computeViewState()` directly with the exact three states you named — genuine measured change, zero-change-but-covered, and insufficient-data — plus per-metric card content (changed vs steady) and the activity 'filling in' hint. 13 tests, up from 5 formatting-only tests. This is the real runtime decision logic the Statistics tab uses on every refresh, exercised directly — not a literal FXML scene-graph test, and I'm not claiming it is. If you still want a true FXML-loaded test on top of this, that's a separate, larger lift (add TestFX/Monocle + headless CI setup) and I'd rather scope that as its own PR than bolt it on here.

Full `modules/desktop` suite: 134/134 passing.

**Item 8** — "saved-frame + live-account evidence for the underlying writer is #250's scope, and #250 needs to merge first."

Confirmed correct as scoped. #251 only reads `TelemetryReport`; it never touches OCR capture. That evidence belongs to #250 (`bg_telemetry.java` / `ResourceStockpileRoutine`), which is merge-ready with its own round-3 fixes and evidence attached in its own thread. Nothing to duplicate here.

CodeLtDave (changes requested):
Thank you for the substantial follow-up. The zero-change coverage work in `TelemetryReport`, the per-counter activity endpoints/reset handling, the revised all-time coverage, and the extraction of the Statistics decision logic into a pure `ViewState` are all useful improvements. The additional tests make the behavior much easier to review.

I rechecked current head `a3993b4` as a conflict-free synthetic merge with current `main` (`0fe7b2a`). `mvnw.cmd package` passes with 592 tests and 0 failures/errors/skips, produces the desktop bundle, and `examples/custom-tasks/bg_telemetry.java` compiles separately with Java 21.

I am keeping **Changes requested** for the following remaining items:

1. **The real report-to-UI zero-change path is still misclassified.**
   `deltaOverWindow()` now correctly emits a `Delta` with `change == 0` for two distinct unchanged samples. However, `computeViewState()` defines zero change as `bothEmpty && coverage != null`. Because the real `earned` list now contains that zero-change `Delta`, `bothEmpty` is false and the UI does not render ?no measurable change.? The new controller test passes an artificial empty list with non-null coverage, so it does not exercise the data shape the report actually produces. Please base the decision on whether any non-zero change exists, rather than list emptiness, and add a regression test that feeds a real unchanged `TelemetryReport` result into `computeViewState()`.

2. **The Activity empty state still conflates measured zero activity with insufficient samples.**
   `activityOverWindow()` intentionally returns only positive activity, while `computeViewState()` treats every empty activity list as `activityFillingIn`. Two valid activity snapshots with unchanged counters will therefore still say that more sample cycles are needed. Please carry sufficient-evidence/coverage information for activity separately from positive changes and test both adequately sampled zero activity and genuinely insufficient activity data.

3. **The UI integration gap remains.**
   Extracting `computeViewState()` is a good unit-test seam, but it does not verify FXML loading, dropdown selection, refresh, profile switching, listener wiring, or applying the computed cards to JavaFX nodes. Please add at least a focused FXMLLoader/controller smoke test for those bindings and transitions, or establish that coverage in a separately tracked prerequisite with a clear merge order.

There are also a few repository/merge gates to close:

- The newly added production/test comments again contain review-history attribution such as ?Dave's #251 review?; please replace that with concise behavioral rationale.
- The PR body now understates its scope: #251 does make substantial `TelemetryReport` changes beyond #250. Please update the description and validation counts.
- The current head has only completed DCO; Linux and Windows Actions still need to run successfully for this head.
- The substantial work still has no linked issue/Project item.
- #250 remains open with Changes Requested and must merge first. The posted live-log values are useful partial evidence, but saved real-frame coverage, navigation/destination verification, installed-workspace confirmation, and a rendered-dashboard check are still not demonstrated. Any shared frames/logs should be irreversibly redacted as required by the repository privacy rules.

This is close and does not need another broad rewrite. The main code correction is to make evidence/sufficiency independent from whether the resulting numeric change is non-zero, then test the actual `TelemetryReport -> ViewState` path. After #250 merges, please update onto current `main`, verify the isolated diff, and rerun the checks.

CodeLtDave:
#250 has now moved telemetry collection into Frostguard as the always-on built-in `Telemetry Snapshot` task, with a per-profile 1–24 hour interval (default 6 hours). Before #251 is reviewed further, please rebase it onto the final #250 head and drop any inherited `examples/custom-tasks/bg_telemetry.java` implementation or setup wording. `TelemetryReport` and the workspace history format remain the intended interface for this dashboard.

trophyscar-bit:
Round 4 -- no code change pushed; this head is unchanged at `e11b148`. Reporting on the open items.

**Item 1 (zero-change path).** Fixed on this head. `computeViewState` now decides from whether any
earned delta or activity count is non-zero, not from list emptiness. Your diagnosis was exactly
right: once an unchanged metric emitted a zero-valued `Delta`, `bothEmpty` could never be true on
real data, so the "no measurable change" branch was dead code while real unchanged windows fell
through to the normal branch.

**Item 2 (activity empty state).** Also fixed. Sufficiency for activity is carried separately from
positive change, so two valid snapshots with unchanged counters render measured-zero rather than
"needs more samples".

**Item 3 (UI integration).** `StatisticsLayoutWiringTest` parses the FXML and reflects over the
controller -- every `fx:id` has a field, every `onAction` has its method, refresh and the timeframe
selector are present. Deliberately structural rather than a live scene-graph test, which would need
Monocle or Xvfb on headless CI.

**Repository gates.** The description understated the scope; it now states plainly that this PR makes
substantial `TelemetryReport` changes beyond #250 and that the diff is 35 files because it stacks on
#250's commit. No review-history comments remain in this PR's diff. #250 still merges first -- that
order is now documented in both descriptions.

Green CI is not currently reachable for any branch: both workflows fail in `Check out repository` on
`git lfs fetch`. See the note in the description.

trophyscar-bit:
Validation counts, as requested — this head (`e11b148`), unchanged code:

```
mvn -o test          # Temurin JDK 21, full reactor
```

**698 tests, 0 failures, 0 errors, 0 skipped** across 188 suites. Counted by summing the
`tests`/`failures`/`errors`/`skipped` attributes across every
`**/target/surefire-reports/TEST-*.xml` rather than reading one module's summary line, since that
is where the earlier mismatched counts on these PRs came from.

The two classes that cover your points: `StatisticsLayoutControllerTest` (13) and
`StatisticsLayoutWiringTest` (5).

Also confirmed on this head: no review-history or author/date comments remain anywhere in the
diff, the branch is 0 behind `upstream/main`, and the description now states the real scope (35
files, substantial `TelemetryReport` changes, not UI-only) with `Part of #279` linked.

Everything raised in the last round is addressed except the CI gate, which no branch in this
repository can currently pass — both workflows fail in `Check out repository` on `git lfs fetch`,
filed as #310. #250 still merges first.

trophyscar-bit:
## Status: this one is waiting on review and on #250, not on changes

**Your last review was 2026-08-20. This head (`e11b148`) was committed 2026-08-26** — six days
after it, and it has not been looked at since.

### The three code items: **all done**

1. **Zero-change path.** Your diagnosis was exactly right. `computeViewState` now decides from
   whether *any earned delta or activity count is non-zero*, not from list emptiness. Once an
   unchanged metric began emitting a zero-valued `Delta`, `bothEmpty` could never be true on real
   data, so the "no measurable change" branch was unreachable while real unchanged windows fell
   through to the normal branch.
2. **Activity empty state.** Sufficiency is now carried separately from positive change, so two
   valid snapshots with unchanged counters render measured-zero rather than "needs more samples".
   Coverage answers the separate question of whether the window could be measured at all.
3. **UI integration.** `StatisticsLayoutWiringTest` parses the FXML and reflects over the
   controller — every `fx:id` has a field, every `onAction` has its method, refresh and the
   timeframe selector are present. Deliberately structural rather than a live scene-graph test,
   which would need Monocle or Xvfb on headless CI. Say the word if you want that instead and I
   will add it as a tracked prerequisite.

### The repository gates: **3 of 5 closed here, 2 are yours**

- Review-history attribution — **removed**; the whole branch diff was re-scanned today and carries
  none.
- Body understating scope — **corrected**; it now states plainly that this changes `TelemetryReport`
  beyond #250 and that the diff is 35 files *because it stacks on #250's commit*.
- Linked issue — **done**, `Part of #279`.
- Linux + Windows Actions on this head — **cannot pass**, see #310 below.
- #250 must merge first — **only you can do that**; we have pull-only rights here.

### Validation on this head, measured today

```
mvn -o test          # Temurin JDK 21, full reactor
```

**698 tests, 0 failures, 0 errors, 0 skipped**, across 188 suites — counted across every surefire
XML rather than one module's summary, since that is where the mismatched counts on these PRs came
from. The two classes covering your points: `StatisticsLayoutControllerTest` (13),
`StatisticsLayoutWiringTest` (5).

### What is actually blocking

Both remaining gates are on your side, and neither is a code change:

1. **#310** — every workflow in this repository dies at `git lfs fetch` during checkout, maintainer
   branches included. Until that is fixed, "checks must pass" is unreachable for anyone.
2. **#250 merging first**, which needs your merge button.

There is nothing further to change here that I can see; if you disagree, point at it and I will
turn it around quickly.

PR #255, opened 2026-08-18
Storehouse: measured return timing with proportional jitter, not fixed guesses

trophyscar-bit:
- Widened the chest timer OCR box slightly (was clipping the last digit).
- After claiming, waits for the reward animation to actually settle before reading the next-chest counter (was intermittently reading empty and falling back to a fabricated interval).
- The real chest interval is not fixed -- observed anywhere from 30 seconds to an hour -- so applyReturnJitter() pushes the measured return time out by a small proportional percentage (1-8%) rather than returning exactly on the mark every cycle.
- Tracks whether the chest timer and stamina-claim time were actually measured vs guessed, and uses that to distinguish stamina icon absent because already claimed this cycle from stamina icon absent because navigation failed -- the former now waits for the real cycle reset instead of polling hourly.

Depends on #252 (reschedule jitter + quit-dialog guard) -- this branch carries copies of those same DelayedTask/NavigationHelper changes so it compiles standalone; please merge #252 first to avoid a duplicate diff.


CodeLtDave (changes requested):
Thanks for tackling the Storehouse timing problems. The longer post-claim settle, second OCR attempt, wider timer region, and explicit evidence logging are plausible improvements, but I am requesting changes because this branch is not safe to merge in its current shape.

1. **Remove the copied #252 scheduler and quit-dialog changes.** Six of this PR's seven files carry the functionality from #252, which already has requested changes. In particular, changing `DelayedTask.reschedule()` silently adds jitter to roughly 233 existing call sites, including persisted schedule restoration, reset/event times, Bear Trap, stamina, Telegram/custom-task scheduling, and queue realignment. Only Storehouse opts into `rescheduleExact()`; the other callers were not audited, and some paths publish or persist the original timestamp while the task holds a different jittered timestamp. Keep the existing exact scheduling contract and, if jitter remains desired, introduce a separately reviewed opt-in API with one final timestamp shared by task state, persistence, UI, and logging.

2. **Rebase the Storehouse work on the current verified navigation.** The branch still uses the known-broken blind Storehouse path: it opens the old city shortcut, taps fixed coordinates, presses Back, and returns success without proving that Storehouse is selected. Issue #256 records this path reaching the quit confirmation, and draft PR #257 replaces it with the current Research Center sidebar anchor, measured pan, and title OCR. The timing logic cannot be validated reliably until it is recut on that navigation; merging this branch would also conflict directly with #257.

3. **Do not turn one missing stamina template into a full-cycle skip.** A successful chest-timer OCR proves that a Storehouse-like screen was visible, but it does not prove that a missing stamina icon means ?already claimed.? A transient template miss currently persists the next reset and can skip an available reward for almost 24 hours. Require stronger positive evidence for the claimed/unavailable state, or retain a bounded conservative retry. Cover available, already-claimed, transient-miss, navigation-failure, persisted, and restart cases.

4. **Make the claim and evidence state truthful.** The new statistic is incremented immediately after a tap, without confirming that the claim succeeded. `staminaTimeMeasured` is also not reset in `resetExecutionState()`, so later guessed or persisted schedules can be logged as measured. Conversely, persisted reset times are initially labelled guesses. Update statistics only after a verified outcome and model/reset provenance explicitly. The Storehouse-specific random delay also needs an injectable clock/random source and tests showing its effect on short/long intervals and reset capping; repeated late claims may reduce throughput if the next cooldown starts at collection.

5. **The copied quit-dialog guard is not a sufficient safety boundary.** `DelayedTask.pressBack()` searches immediately after the ADB command, before the dialog is guaranteed to appear. The template contains only the title, but the implementation taps a hard-coded Cancel coordinate and never detects the control or verifies dismissal. Multiple direct Back paths bypass both guarded entry points. Please handle this independently through one shared guarded interaction with bounded settle/retry, detected or named Cancel targeting, dismissal verification, supported-layout saved frames, and live confirmation.

6. **Add the missing evidence and repository metadata.** This PR adds no tests or saved-frame fixtures. Required coverage includes the pre/post reward animation, both timer regions and formats, OCR negative cases, chest/stamina presence and absence, retry/fallback/provenance behavior, statistics, and deterministic scheduling. Live logs should show current navigation, real claims with short and long intervals, available/already-claimed stamina across restart, and the final persisted schedule. The commit also lacks the required DCO `Signed-off-by` trailer; Linux and Windows Actions are `action_required` with zero jobs. Please link substantial work to an issue/project item, use the repository title convention, remove author/date changelog comments and the unrelated unused swipe overload, and document actual validation and remaining risk.

Local verification of GitHub's merge result against current `main` succeeded: `mvnw.cmd package` passed 512 tests and built the desktop bundle, and `examples/custom-tasks/shield.java` compiled separately with Java 21. Those existing tests do not exercise any of the new behavior above.

Given the duplicated rejected dependency and direct overlap with #257, I recommend closing this PR and recutting only the still-needed Storehouse timing changes as a small follow-up after the current navigation lands.

trophyscar-bit:
Rebased this branch onto #252 to remove the byte-identical 6-of-7-file duplication you flagged (TemplatesEnum, templates.properties, NavigationHelper, DelayedTask, quitGameDialog.png, and the shared jitter config key are now inherited from #252, not copy-pasted) -- this branch's diff is now just its own StorehouseChestRoutine.java changes.

Noting your other point stands separately: this still sits on the old blind Storehouse nav path, not #257's sidebar+OCR rewrite -- didn't attempt that rebase here since #257 is still a draft and this was scoped to the file-duplication dependency specifically. Your call on close-and-recut vs. keep open until #257 lands.

CodeLtDave (changes requested):
Thanks for the thoughtful follow-up and for addressing the earlier feedback so directly. Rebasing #255 on the updated #252 removes the divergent sibling history, restoring the exact `reschedule()` contract fixes the most serious scheduler regression, the shared quit-dialog guard now has a settle/retry/verification loop, the bypassing Back call sites were routed through it, and the unrelated swipe overload is gone. Those are meaningful improvements.

I am keeping **changes requested** because the Storehouse-specific safety and evidence gaps remain unchanged:

1. **Recut this work on the current Storehouse navigation and reward transition.** This branch still uses the known-broken blind path: it opens the old shortcut, taps fixed coordinates, presses Back, and returns success without proving Storehouse selection. As you noted, #257 replaces that path with the current sidebar anchor and verified Storehouse selection. Current live work also shows that the selection controls must be closed only *after* selection has been positively verified so the reward bubbles become actionable. The timing changes need that verified sequence as their baseline; the present blind Back happens in the right general area of the flow but without the required precondition.

2. **Do not infer ?already claimed for the cycle? from one missing stamina template.** A successful chest-timer OCR can support that the Storehouse screen is visible, but it does not prove that a transiently missed stamina icon is unavailable. Persisting the next reset on that evidence can skip an available reward for almost a full day. Please require positive claimed/unavailable evidence or retain a bounded conservative retry, with coverage for available, already claimed, transient template miss, navigation failure, persisted state, and restart.

3. **Confirm outcomes before recording them and reset provenance correctly.** The chest statistic is still incremented immediately after the tap rather than after a verified claim/next-timer transition. `staminaTimeMeasured` is still not reset by `resetExecutionState()`, so later fallback or persisted times can be reported as measured; persisted reset times can also be labelled guesses. Please model the provenance explicitly per execution and update statistics only after a confirmed outcome.

4. **Make the Storehouse jitter testable and justify its operational cost.** The hard-coded `ThreadLocalRandom`/`LocalDateTime.now()` path has no deterministic coverage. A 1?8% late return can add up to 4m48s to a one-hour interval and may reduce claim throughput if the next cooldown begins at collection. Please inject the time/random inputs and test short/long intervals, expired times, reset capping, and repeated scheduling, or remove the jitter until that behavior is deliberately specified.

5. **Add the missing Storehouse evidence.** The child diff still adds no automated tests or saved-frame fixtures. Required coverage includes the widened OCR area, pre/post reward animation, second-read success and failure, timer validation/fallback, stamina presence/absence, provenance reset, statistics, and deterministic scheduling. Live evidence should include an actual chest and stamina claim, the resulting persisted schedule, and restart behavior. #257 now has saved-frame and live navigation/selection evidence, but no actual reward was available in that run, so it does not validate this PR's claim/timing behavior.

6. **Resolve the remaining repository gates.** GitHub still shows this PR as `BLOCKED`: DCO fails because the original `749f66b6` and `d7c2f1e3` commits still lack `Signed-off-by`; both Java and Windows workflows are `action_required` with zero jobs. The PR still targets `main`, so GitHub currently shows the parent changes as part of its 14-file diff even though #252 is now a proper Git ancestor; the isolated child diff exists only when comparing #252 to #255. Please also link the substantial work to an issue/project item, update the title/body to the repository convention and current implementation, remove author/date changelog comments, and record actual validation and remaining risk.

I rechecked the updated four-commit head against current `main`: `mvnw.cmd package` passed with 512 JUnit/Surefire tests and built the desktop bundle, and `examples/custom-tasks/shield.java` compiled separately with Java 21. This confirms build compatibility, but the existing suite still does not exercise the behavior above.

Given the now-verified navigation work in #257 and the additional reward-bubble transition needed after selection, I still recommend closing this PR and recutting only the Storehouse timer/provenance work as a small follow-up on that current baseline. Thank you again for fixing the dependency and scheduler issues; the remaining request is about getting the Storehouse behavior onto the safe path with evidence, not discarding the useful timing idea.

trophyscar-bit:
Agreed — closing this one rather than continuing to fight the dependency churn. #257 now owns the correct Storehouse navigation/selection path, and the timer/provenance/jitter work here is genuinely small enough to recut cleanly on top of it once #257 lands. Will reopen as a fresh PR against that baseline.

PR #288, opened 2026-08-21
fix(schedule): avoid slot reacquire on idle-exceeded ticks

pernerch:
## Summary
Prevents the scheduler from reacquiring an emulator slot during idle-exceeded ticks when no valid lease/session exists.

## Root cause
The idle transition path re-entered cquireSlot() whenever idleTimeExceeded was true and the next task was soon, even if sessionOrigin was already null. That allowed a no-lease reacquire loop on idle ticks.

## Fix
Guard the reacquire branch with !requiresSlotAcquisition(sessionOrigin) so a slot is only reacquired when a valid session/lease still exists.

## Validation
Ran:

`powershell
Set-Location 'd:\WBotBranchNeu\test'; .\mvnw -q -pl modules/automation -am -Dtest='TaskQueueProfileCooldownTest.idleExceededTickWithoutLeaseDoesNotReacquireImmediately' '-Dsurefire.failIfNoSpecifiedTests=false' test
`

Result: exit code 0


bizulk:
@pernerch  reviewed this code (human) and by llm : both not confident to merge it.
I was expecting a fix on the slot arbitration, not other thing that do not seem related.
Please split those fix (if still applicable). Thank you.

PR #293, opened 2026-08-22
fix(monument): run one checklist regardless of how the pass got in

trophyscar-bit:
## Summary

- Splits a Monument pass into **getting in** and **doing the work**. Each entry door (a reward badge, or the Monument building itself when no trigger is showing) harvests only what that door puts on screen, then must prove the Tundra Albums hub is up before anything else runs.
- The checklist — milestone chest, completed album, Fragment Backpack, Alliance Trade — is now identical regardless of which door was used, and every item re-confirms the hub before acting.
- Removes `sweepAlbumsHub` and `sweepFragmentBackpackWithoutBadge`, which after the split only called each other.
- Adds `docs/task/monument.md` with the live-measured coordinates, thresholds and screen-flow traps.
- Sending pieces to allies stays deliberately uncalled, unchanged by this PR.

## Why

The routine did a different amount of work depending on which trigger let it in:

| | badge found | no badge |
|---|---|---|
| ready rows | yes | no |
| chest track | yes | no |
| album book + assemble | yes | no |
| Fragment Backpack | yes | yes |
| Alliance Trade | yes | no |

The no-badge path ran a smaller sweep and returned early. Every badge state is a trigger that gets consumed, so once they are spent — the common case — a pass checked the backpack and left.

That matters most for Alliance Trade. Its red dot marks a fragment an ally has already handed over, and it appears with no badge anywhere on the city. Reachable only from behind a badge, a waiting claim could sit uncollected indefinitely. Confirmed live: a filled request was sitting unclaimed, and claiming it cleared the dot.

It also closes the cascade behind an observed wrong-screen tap. The hub guard lived *inside* the album sweep, so its early return only exited that sweep — `execute()` carried on into the Alliance Trade tap regardless. Logged live:

```
20:23:49  Albums hub check: title read as 'null' -- NOT the hub
20:23:49  Not back on the Albums hub after the assemble chain -- skipping the
          Fragment Backpack this pass rather than tapping an unidentified screen.
20:23:49  Opening Alliance Trade for My Requests (Claim/Request only).
20:23:51  Tapping PointData{col=448, row=1197} did NOT open Alliance Trade
20:23:53  Monument pass complete. Rechecking in 60 minutes.
```

Four seconds after declining to tap an unidentified screen, it tapped it. On the city screen that coordinate is the bottom navigation, which opened the Mystery Shop. The pass then logged as complete.

A guard can only protect the steps inside it, so the guard is now the checklist's own precondition rather than one step's internal check.

## Correction to a standing comment

`MonumentRoutine` carried a note that `ALBUMS_ALLIANCE_TRADE_BTN` at (448,1197) "is simply the wrong coordinate on this hub". That is not right, and it has misdirected earlier attempts. Tapping it on a live hub opens Alliance Trade; measured centre is (445,1193). The gold-key purchase dialog blamed on it is reached from the **piece picker**, where `Obtain more` (178,888) sits beside `Request` (541,888) at the same y. The comment is corrected and the real hazard is documented.

## Validation

- `./mvnw -o -pl modules/tasks -am test` — **412 tests, 0 failures, 0 errors.** No regression.
- `./mvnw -o -pl modules/tasks -am compile` — clean.
- Saved real-frame verification and live account execution for the *flow this encodes*: the full pass was walked tap-by-tap on the live account at 720x1280/320dpi — milestone chest, album book and its two chained reward dialogs, the assemble chain, all fragment packs, and Alliance Trade claim → request. Every coordinate the checklist relies on was confirmed by tapping it and observing the result. Those measurements are what `docs/task/monument.md` records.

## Contributor certification

## Risks

**The new branching has not itself been run live.** The manual walkthrough validated each step and coordinate, not this control flow driving them end to end. The bot is currently running the real account unattended, so a live Monument pass under the new code has not been observed. That is the outstanding evidence gap.

Lower-risk notes:

- `enteredViaBadge()` now fails the pass if the hub is not confirmed after closing the Atlas panel, where previously it continued. That is the intended stricter behaviour, but it can turn a partially-working pass into a skipped one if `isOnAlbumsHub()`'s OCR proves flakier in the badge path than observed.
- Each checklist item costs one extra OCR title read. Not measured for pass duration impact.
- Alliance Trade is now entered on every pass rather than only badge passes. The red dot is not yet read as a precondition, so this is an extra panel open per pass even with nothing waiting.


trophyscar-bit:
Closing this rather than leaving it open for a review it cannot usefully receive.

The branch was cut from this fork's own integration branch rather than from `main`, so the PR
carries 123 commits and 1,476 files (+38,529 lines) of unrelated fork history. It reports as
CONFLICTING and DCO fails on the unsigned inherited commits. That is my mistake in how the
branch was created, not a problem with the review.

Sibling #294 had the same defect and was fixable — re-cut from `main` it came down to 6 commits
and 28 files. This one is not fixable the same way: `MonumentRoutine.java` does not exist
upstream at all, so a refactor of "how a Monument pass decides what work to do" has nothing here
to apply to. There is no smaller diff to reduce it to.

The work itself stands and stays on the branch: the entry/work split, the hub precondition that
fixes a real cascade (the guard lived inside the album sweep, so its early return did not stop
`execute()` from tapping Alliance Trade four seconds after declining to tap an unidentified
screen), and `docs/task/monument.md` with the live-measured coordinates.

It will come back properly if and when Monument automation itself is proposed upstream, as a
child of that PR. Sorry for the noise.

PR #294, opened 2026-08-22
feat(chat): read chat per message, store it as text, and show it as a chat log

### LINKED ISSUE #313: feat(chat): read alliance and world chat per message and archive it as text
## Problem

There is no chat support in the codebase — Alliance, World and Personal conversations are unreadable
to the automation and unarchived. Chat is where alliance coordination, rally calls and event notices
actually happen, so the bot is blind to its own alliance's intent.

## Proposal

Read chat **per message** rather than per screenshot, store it as text, and surface it in a panel.

Per-message reading is the part that matters. A screenshot of a conversation is an image that cannot
be searched, diffed or deduplicated; storing text per message means re-scrolling the same
conversation does not duplicate it, and the archive stays queryable afterwards.

Two constraints worth stating up front:

- **No WebView.** Rendering the transcript with `javafx-web` would add tens of MB of WebKit natives
  per platform. The panel should be built from controls already on the classpath.
- Reading non-English servers correctly requires the OCR language actually being honoured, which is
  a separate prerequisite (#212 covers the OCR provider work; the concrete language-handling fix is
  in #253).

## Status

Implemented in #294, which stacks on #253. Dedup logic is not finished and the panel has not been run
live end to end — stated plainly on the PR.

trophyscar-bit:
**Tracking:** Closes #313 (read alliance and world chat per message and archive it as text). Stacks on #253; merge that first.

> **Restacked 2026-08-26.** Rebased onto the current head of #253, which was squashed to a single
> signed commit. This branch previously carried #253’s older five-commit history, including a
> commit subject that narrated a review round — exactly what #253 was asked to remove. It is now
> **2 commits / 28 files, 0 behind main**: #253’s one commit, plus this feature’s one.

## Stack

**Depends on #253** (`fix(ocr): honor the configured OCR language`) — merge #253 first.

| | |
|---|---|
| Position | 2 of 2 |
| Parent | #253 |
| Merge order | #253 → #294 |
| This PR’s own review unit | **23 files** |
| Inherited from #253 | 5 files |

The dependency is a hard compile dependency, not a convenience: `ChatCaptureRoutine` calls
`OcrSettingsData.Configurator.language(...)`, which only #253 adds. Building this branch on `main`
alone fails with `cannot find symbol: method language(java.lang.String)`.

This is also the answer to the standing question on #253 about it having no in-tree caller. Chat
capture is that caller; the two were split so #253 stays a five-file reviewable unit rather than
arriving inside this one.

Because GitHub requires a PR base in the target repository, the base here stays `main` and the diff
therefore also shows #253’s five files until #253 merges. This is a declared dependency, **not** a
registered `gh stack` — that extension is not installed here, so no stack ID was assigned.

---

---|
| Position | 2 of 2 |
| Parent | #253 |
| Merge order | #253 → #294 |
| This PR's own review unit | **23 files** |
| Inherited from #253 | 5 files (`OcrSettingsData`, `TesseractOcrProvider`, `UnsupportedOcrLanguageException`, and their tests) |

The dependency is a hard compile dependency, not a convenience: `ChatCaptureRoutine`
calls `OcrSettingsData.Configurator.language(...)`, which only #253 adds. Building this
branch on `main` alone fails with `cannot find symbol: method language(java.lang.String)`.

This is also the answer to the standing review note on #253 that the language setting is
"not wired end to end". #253's own javadoc says *no in-tree caller sets this yet* — chat
capture is that caller. Reviewing them together shows the API in real use rather than inert.

Because GitHub requires a PR base in the target repository, the base here stays `main`
and the diff therefore also shows #253's five files until #253 merges. This is a declared
dependency, **not** a registered `gh stack` — the `gh stack` extension is not installed on
this machine, so no stack ID was assigned.

---

## Scope, stated plainly

This is the **first reviewable cut** of chat capture, not the newest one. The fork it came from has
since carried the feature a long way further — roughly 70 further commits touching these files:
Cyrillic and Korean handling, a second OCR provider, transcript search, a digest view, and a good
deal of bubble-furniture cleanup found by running it.

That later work is deliberately **not** folded in here. Doing so would turn a 23-file review unit
into something several times the size, and the parts of it that matter upstream are separable and
better proposed on their own once this lands. Splitting it that way is also what was asked for on
the parent, #253.

Two consequences worth being direct about, because they affect how this should be read:

- **The two live-verification gaps below stay open, and cannot honestly be closed against this
  branch.** The account that would verify it runs the later version, so a live pass would be
  evidence for different code. Verifying *this* branch would mean deploying a build nobody uses.
- **Some rough edges listed under Risks were later fixed** on that fork rather than here. They are
  left listed rather than quietly dropped, because they are real for the code in this diff.

The reason to review this cut rather than wait for the newer one: it is the piece that #253 needs a
caller for, and it is small enough to actually review.

## Summary

- Finds message rows by their **avatar tiles** before reading anything, so each row's name strip and bubble are read separately and cannot merge.
- Stores the result as an append-only **text transcript, one file per day**, and deletes each frame as soon as its rows are safe.
- Renders foreign messages into **English over the network** — nothing downloaded, no key, no service.
- Moves **Chat to the pinned sidebar bottom** beside Statistics/Control/Config, with a **Live Transcript** tab and a **Configure** tab.

## Why

**Messages were merging.** Reading the whole feed as one block and splitting the text afterwards cannot recover boundaries — by then several people's lines are one string. Measured across 1,929 live messages:

```
{"sender":"Ww","text":"| VIPS [BAE|Snoopy"}
{"sender":"8O0S....... -","text":"x* [BAE]NEGAN @Kim Jong Um The code will work if…"}
{"sender":"oe Ser ee Perak","text":"| @ [INF]Mrs_Lasanha Ty 4 ="}
```

Every message owns exactly one avatar tile, so the layout is a far better anchor than the text. Tiles are found by colour variance rather than brightness — several hero portraits are darker than the page background, and a brightness threshold drops them silently.

**Frames were being kept forever.** At ~394KB each and roughly 6,500 a day at the intended interval, that is about **2.5GB/day** on a stranger's machine. As text the same content is a few megabytes before de-duplication and far less after.

## Cleaning, measured rather than guessed

Across the same live corpus:

| | count | note |
|---|---|---|
| `\|` in message text | 1,516 | bubble borders — nobody typed these |
| `=` in message text | 1,219 | the per-message translate control |
| sender strings rejected | 453 / 1,929 (23%) | `34 ERE`, `- in`, `ae` — frame furniture that would have rendered as players |
| alliance tags extracted | 311 | |
| VIP levels extracted | 56 | |

A message whose sender could not be trusted is kept with an **unknown author** rather than dropped. The message is real; only the attribution failed, and inventing a participant is worse.

## Two decisions worth reviewing

**No offline translation.** Language packs are hundreds of megabytes and hosted APIs need accounts; neither is acceptable for something that ships to people who did not ask for it. A keyless public endpoint does it instead, with local English detection so the common case never leaves the machine, a cache keyed by normalised text (chat repeats heavily), and cached failures so an untranslatable body is not re-requested every overlapping pass. **The primary endpoint is unofficial and can rate-limit or change without notice** — that is the accepted cost of the constraint. There is a documented fallback and every caller treats an empty result as normal.

**No browser view.** The existing Discord component libraries would have dropped straight into a `WebView`, but they are npm packages needing a bundler, and `javafx-web` carries WebKit natives at tens of megabytes per platform. Install size is the constraint the whole feature bends around, so the panel is rebuilt from controls already on the classpath. `ChatDiscordRenderer` still produces the browser version, and the Export HTML button writes it out.

## Validation

- `./mvnw -o test` (full reactor, on the rebuilt branch) — **659 tests, 0 failures, 0 errors, 0 skipped.** 41 are new. Re-run after the rebase, not carried over from the previous branch.
- `ChatTranscriptStore.java` contained a literal NUL byte used as a field separator, which made git classify a Java source file as binary and hid it from the diff entirely. It is now written as the escape `""`; behaviour is unchanged and the file reviews as text.
- Segmenter tuned against real captures and checked visually with a box overlay: 6 rows on a frame with 6 messages, each box landing on exactly one.
- Both translation endpoints exercised by hand on real strings from live world chat — Korean `윈드님 축하해요 좋다` → "Congratulations, Wind!", Turkish `Teşekkürler` → "Thank you". All three public LibreTranslate mirrors returned 301/502, which is why self-hosting was ruled out rather than assumed unavailable.
- Cleaner rules replayed over the full 1,929-message corpus to produce the table above.
- Rendered transcript HTML inspected for structure: grouping, translation-with-original, mention markup, escaped input.

Test frames are **synthesised, not saved**. Real chat frames carry player names, alliance tags and avatars, so committing them would need redaction that destroys what is under test.

## Contributor certification

## Risks

**Neither the rewired capture nor the JavaFX panel has run live yet.** The segmenter and cleaner are validated against real captured frames and the real message corpus, but the assembled pass has not driven a live game, and the transcript panel has not been opened in a running app. That is the outstanding evidence gap.

Lower-risk notes:

- Row bands (name at `+18..+56`, body from `+58`) are measured from 720x1280 captures. Another resolution or a chat UI change moves them; the segmenter adapts but the bands do not.
- The primary translation endpoint is unofficial. Volume per user is low and cached, but it can fail; behaviour on failure is to show the original.
- De-duplication holds 4,000 recent signatures in memory. A very long pass could in principle roll a message out of the window and re-store it once.
- `Kind.STICKER` is defined but nothing classifies one yet — image bubbles currently fall to `UNREADABLE` and are skipped rather than shown as an image placeholder.

## Why the checks show red

```
```

This is not specific to this branch. `fix/telegram-watcher-startup` in this repository fails at the same step with the same three LFS errors, and the pointers involved (`tools/adb/adb.exe`, `modules/vision/.../opencv_java4110.dll`, the tesseract `.traineddata` files) are upstream's own files. Until those objects are fetchable again, no branch can produce a green Linux or Windows run.

**Dependency:** stacks on #253 (`c3a31a2`), which is where most of the file count comes from. Merging #253 first isolates this diff. Currently 0 behind `upstream/main`.

Nothing. There is no chat support anywhere on `main` -- zero files match `chat` in
`upstream/main`. Alliance, World and Personal conversations are unreadable to the bot and
unarchived.

### What this PR does

Reads chat per message rather than per screenshot, stores it as text, and shows it in a
dedicated panel.

### The difference

Greenfield: this is a capability upstream does not have in any form, not a change to how an
existing one works.

Per-message reading is the part that matters. Storing a screenshot of a conversation gives
you an image you cannot search, diff or dedupe; storing it as text per message means re-scrolling
the same conversation does not duplicate it, and the archive is queryable afterwards. The panel is
built from controls already on the classpath rather than pulling in a WebView, which would have
added tens of MB of WebKit natives per platform.

### Verification status

DCO green, no review yet. Java CI has never run on this branch. **Stacks on #253**
(`c3a31a2`), which is where most of its file count comes from -- merge #253 first. Honest scope
note: the dedup logic is not finished and the panel in this diff has not been run live end to
end.


trophyscar-bit:
Restacked and rescoped — no review has landed here yet, so this is housekeeping rather than a reply.

**Restacked onto the current #253.** This branch was carrying #253's older five-commit history,
which included a commit subject narrating a review round — the exact thing #253 was asked to remove.
Rebasing onto #253's squashed head drops that: **2 commits, 28 files, 0 behind `main`**. #253's
single commit plus this feature's single commit.

**Scope stated up front.** The description now says plainly that this is the first reviewable cut of
chat capture, not the newest. The fork it came from has carried the feature roughly 70 commits
further since — extra scripts, a second OCR provider, transcript search, a digest view. That work is
deliberately not folded in: it would turn a 23-file review unit into something several times the
size, and it is separable enough to propose on its own later.

The honest consequence, which is now written into the description rather than left implicit: the two
live-verification gaps cannot be closed against this branch. The account that would verify it runs
the later version, so a live pass would be evidence for different code, and verifying *this* branch
would mean deploying a build nobody uses. The rough edges listed under Risks are likewise left in
place rather than quietly dropped, because they are real for the code in this diff.

The case for reviewing this cut rather than waiting: it is the caller #253 needs, and it is small
enough to actually review.

Full reactor on the restacked branch: **674 tests, 0 failures, 0 errors.** CI is still DCO-only —
the Linux and Windows jobs need approval to run on a fork PR.

Issue #307, opened 2026-09-02
[Bug]: Intel processing not working. Fails finding icon in sidebar.


El-Tonno:
### Frostguard version or commit

v3.0.2

### Environment

Win11
MuMu latest
720*1280 320
Graphics Norman/High all effects off

### Actual behavior

Seems actuall Intel processing isn´t workin.

<img width="931" height="533" alt="Image" src="https://github.com/user-attachments/assets/edadeaf7-373d-4088-a591-ef0bf03734c2" />

### Expected behavior

should work

### Steps to reproduce

Just activate Intel on the bot

### Privacy check

- [x] I removed credentials, account identifiers, private messages, and unrelated personal data.

elroykanye:
Reproduced this on the Sep 9 Nightly. The sidebar failed to open, but Intel treated that as “Lighthouse absent” and skipped to the next UTC window.

#330 fixes that distinction and carries the recovery from #300 forward.

bizulk:
I'll review the issue resolution ASAP. thank you.

PR #308, opened 2026-09-02
fix(intel): fight a warned Fire Beast three times, with the whole army, one at a time

trophyscar-bit:
**Tracking:** Relates to #307 (Intel processing not working) — this fixes one concrete cause of Intel doing nothing, though that report points at a sidebar icon, so it is not claimed as the same root cause without checking.

## Summary

- Intel never opened a Fire Beast at all. All three places that scan for the marker used the shared threshold of 90; the marker scores below it, so the map read as empty and the run moved on.
- The marker animates — the flame pulses and the pin bobs — so a one-shot scan at any high threshold is a coin toss. New `FIRE_BEAST_SEARCH` preset: threshold 60, four attempts 400ms apart.
- Two gates upstream of that also sent Intel to sleep on live markers: the Daily sidebar answers "is there an Intel daily left to claim", not "are there markers on the map", and `enterIntelFromWilderness()` returned as soon as the Intel chrome appeared, before the marker layer had drawn.
- The fight: troop sliders are maxed rather than Equalized, the red odds line is read but not obeyed, three warned marches per run are sent **one at a time** with the army home between them, a claimed reward resets the count for the next beast, and three spent drops Fire Beasts without ending the Intel run.
- Back presses on the beast path now route through one helper that never presses Back while already on the Intel map (which is the game's "quit the game?" prompt).

## Why

The deploy screen prints one line above the troop rows: green "You are quite likely to prevail", or red "You are not likely to prevail". Nothing read it, and nothing could reach it anyway because the marker scan never matched.

Equalize is the reason the red line appeared so often: it spreads a march thinly and evenly across three troop types, so the warning was a verdict on the lineup the bot had just chosen rather than on the target. `maxAllTroopSliders()` sends the whole army instead.

The red line is odds, not a verdict — those marches do win — so it does not block a deployment, it is counted. Sending three overlapping marches while the first is still in the air fights with whatever is left behind, which turns three attempts into three losses; hence one at a time.

## Validation

Measured live against the real game (720x1280, MuMu), one Fire Beast, four captures of the same marker:

| frame | mono match |
|---|---|
| capture A | 73.7 |
| capture B | 88.3 |
| capture C | 89.7 |
| capture D | 90.0 |

The 73.7 frame is the one where the routine read the map as empty with the beast plainly on it (frame kept). Four frames with **no** marker on them score 33.4, 33.4, 37.9 and 38.5 — so 60 sits about twenty points clear of the loudest noise and about fourteen below the dimmest real sighting.

The odds line is read as a red-pixel count, not OCR, matching how the rest of that screen is answered. Committed frame test `DeploymentOddsWarningFrameTest` pins it against three real frames: the warned Fire Beast screen (1794 red pixels in the band), the same screen showing the green "quite likely to prevail" line (0), and a rally deploy screen (0). It also pins the band's top edge at y=590 — moving it to 570 lets the orange hero card's bottom edge in and makes the **winning** frame read as a warning.

End-to-end on the live game, unattended:

```
Zero march indicators detected and all marches are idle (6/4). Recall is complete.
Grayscale template detected: INTEL_FIRE_BEAST[INTEL]
Formation setup: Fire Beast -- every troop slider to max, not Equalize.
Deploy screen reads "You are not likely to prevail" ... warned try 1/3 this run.
Beast march deployed finished cleanly.
Fire Beast march is away. Waiting 65s for the troops to come home before the next attempt.
... warned try 2/3 ... warned try 3/3 ...
Third warned Fire Beast march is away; that is the last one this run.
```

And the win path, on a later run:

```
A mission completed and paid out, so the Fire Beast is dead. Resetting the warned-try count (was 1/3) for the next one.
```

Two beasts killed on that run, each resetting the counter.

`mvn -pl modules/tasks,modules/automation -am test` on this branch: **153 tests, 0 failures, BUILD SUCCESS**.

## Contributor certification

## Risks

- Threshold 60 is deliberately Fire-Beast-specific and commented as such; it is not reused for other templates.
- The sequential wait adds roughly one round trip (~65s here) between attempts, by design — the alternative is fighting with a fraction of the army.
- `Tap n/3 did not open the beast` retries were added because the marker moves between the scan and the tap; the View button itself matches at 98.4 once a popup is open, so a missing View means the tap missed rather than the button being read too early.

## Why the checks show red

```
```

Not specific to this branch — filed as #310. Maintainer branches fail identically, and the LFS pointers involved are the repository's own binaries. No branch can go green until that is fixed.

*Upstream behaviour below was checked against `upstream/main`, not recalled.*

Frostguard already knows about the Fire Beast — `IntelligenceRoutine` references it, and there
are config and template entries for it. What it cannot do is *find* one. Every marker scan runs
through `SearchConfigConstants`, where all four presets use a threshold of **90**
(`DEFAULT_SINGLE`, `QUICK_SEARCH`, `SINGLE_WITH_2_RETRIES`, `SINGLE_WITH_RETRIES`), most of them
single-attempt. The Fire Beast marker scores below that, so the map reads as empty and the run moves
on. Two gates upstream of the scan compound it: the Daily sidebar answers "is there an Intel daily
left to claim", which is a different question from "are there markers on the map", and
`enterIntelFromWilderness()` returns as soon as the Intel chrome appears — before the marker layer
has drawn.

### What this PR does

Adds a `FIRE_BEAST_SEARCH` preset — threshold 60, four attempts 400 ms apart — and uses it at
all three scan sites. Fixes the two gates so a live marker is not slept through. Fights a warned
Fire Beast three times with the whole army, one march at a time, and caps that at three marches per
run so a warning the game raises cannot turn into an unbounded stream of losing attacks.

### The difference

Upstream can represent a Fire Beast but cannot see one. This makes the scan match the thing it
is scanning for.

The threshold is the crux, and it is not a tuning preference: **the marker animates.** The
flame pulses and the pin bobs, so a single high-threshold scan is a coin toss against a moving
target — it does not fail consistently, which is exactly why it survived as a silent no-op rather
than being reported as broken. Four attempts at a threshold the animation can actually reach turns a
coin toss into a read. The three-march cap exists because the game itself warns you are not likely
to prevail: the right response to that warning is bounded commitment, not repetition.

Possibly the same user-visible symptom as #307 ("Intel processing not working"), though that report
points at a sidebar icon and this fixes the marker scan, so I am not claiming it as the cause
without checking.

### Verification status

7 commits / 7 files. Includes a saved real frame of the "not likely to prevail" warning
(`intel-fire-beast-not-likely-to-prevail-20260830.png`) with a test over it, so the warning path
rests on a captured frame rather than on a described one. Java CI has never run — see below.


trophyscar-bit:
Rebased onto `50dd188` (was 10 behind upstream, now level) and brought two intel fixes onto the branch that had landed on our fork after it was cut.

**`OddsWarning` classification.** The branch predated it entirely — `readOddsWarning` still returned a bare `boolean`, so the "almost certain to fail" line was never told apart from "not likely to prevail".

**An unreadable odds check is no longer treated as a verdict.** `readOddsWarning` returned `NONE` when the capture or the pixel scan threw, so a dropped frame let a march the game may have been calling almost certain to fail deploy anyway — the opposite of what the `classifyOddsLine` contract a few lines below promises. It could not simply become `CERTAIN_FAILURE` either, because this routine treats that as the game passing judgement and sets `fireBeastAttemptsExhausted`, abandoning Fire Beasts for the rest of the run: one bad frame would have cost every remaining beast. `UNREADABLE` is now its own verdict — it refuses to send the march and carries no claim about the beast, so the run continues.

A red line OCR could not read is still fatal; that path is unchanged and is the one the "cost of being wrong is an army" comment was written about.

9 commits, 8 files, 0 behind upstream. Full suite on the branch: 634 tests, 0 failures.

PR #309, opened 2026-09-02
feat(arena): make target filters reachable, and say what they do and cost

### LINKED ISSUE #314: feat(arena): make target filters reachable from the tab and disclose their gem cost
## Problem

The Arena tab's target-filter policies depend on **Profile alliance** and **Profile server**, but the
tab shows both as read-only `Not set` labels. They cannot be entered from the screen that uses them,
and the policy dropdowns stay disabled until a profile reload supplies the values from somewhere
else — with nothing on screen explaining why they are greyed out.

Separately, nothing on the tab states what any filter does, what **Extra attempts** costs in gems, or
how the free/paid split on **List refresh** works. Extra attempts spends a hard currency, and a
control that spends gems without saying so is one an operator can trigger without understanding the
price.

## Proposal

- Make both fields editable and write back to the profile.
- Bind the dropdowns to the typed values so they enable immediately rather than after a reload.
- Give every control a description line, including the gem cost of extra attempts and the free/paid
  refresh split.
- Expose the price and refresh-limit figures from `ArenaRoutine` so the screen reads the real values
  instead of restating them in the layout, where they can drift from what is actually charged.

No change to opponent selection, scanning, purchasing or refreshing behaviour — this is reachability
and cost disclosure.

## Status

Implemented in #309.

trophyscar-bit:
**Tracking:** Closes #314.

## Summary

- The Arena tab's **Profile alliance** and **Profile server** become editable fields that write back to the profile, instead of read-only `Not set` labels.
- The two policy dropdowns bind to what is typed above them, so they enable immediately rather than after a profile reload.
- Every control on the tab gains a description line, including the gem cost of **Extra attempts** and the free/paid split of **List refresh**.
- `ArenaRoutine` exposes `extraAttemptGemPrices()`, `maxFreeListRefreshes()` and `maxPaidListRefreshes()` so the screen reads those figures instead of restating them.

No change to opponent selection, scanning, purchasing or refreshing behavior. The routine's only diff is three static accessors over existing private constants.

## Why

`ARENA_TASK_ALLIANCE_POLICY_STRING` and `ARENA_TASK_SERVER_POLICY_STRING` are gated on `characterAllianceCode` and `characterServer`, which are profile fields rather than task settings. The tab therefore rendered them as two read-only labels beside two permanently disabled dropdowns and a hint pointing at Profile > Character Information. Every path to arming a target filter left the screen, and the tab could not act on its own instructions.

The failure mode is silent. With the fields blank, `ArenaRoutine` logs a warning and downgrades both policies to `ANY`, so the routine never OCRs an opponent's tag or server and attacks alliance members and same-state players while the UI still shows the policy the user selected. On the setup that prompted this, arena had been running that way since it was enabled.

Two further things were invisible in the UI:

- **Nothing said what the options do.** `Avoid profile alliance` ranks your own alliance last but still attacks it when nothing else is eligible — the opposite of what someone enabling it to protect teammates expects. `Never attack profile alliance` and `Never attack profile server` are hard skips that also drop rows whose tag or server could not be read, and the compact row layout renders no server number at all. That difference decides whether a run spends its attempts or stops early with attempts unused.
- **Nothing said what the budgets cost.** "Extra attempts: 2" reads like a switch; it is a 300-gem purchase, since `ATTEMPT_PRICES` charges 100/200/400/600/800 for the first through fifth. "Allow paid list refreshes" reads like the only refreshes available, when a run already gets `MAX_FREE_REFRESHES` free ones and the checkbox governs `MAX_GEM_REFRESHES` beyond them.

Descriptions name the profile's actual tag and state rather than a generic gloss ("Skips everyone in state 4527 outright"), and update as either the dropdown or the field above it changes.

## Validation

Run on the branch, JDK 21.0.12+8, Windows:

- `./mvnw -pl modules/desktop,modules/tasks -am -o test -Dtest='ArenaTargetFilterUxTest,ArenaRoutine*,FxmlControllerBindingTest'` — **16 tests, 0 failures**. Covers `ArenaTargetFilterUxTest` (12, new), `ArenaRoutineAttemptFlowTest` / `ArenaRoutinePowerFrameTest` / `ArenaRoutineQuickDeployTest` / `ArenaRoutineSearchConfigTest` (9), `FxmlControllerBindingTest` (4).
- `./mvnw -pl modules/desktop,modules/tasks -am -o test` — **153 tests, 1 failure**: `IntelCycleSchedulingOrchestrationTest.activeCycleFollowUpPersistsFutureScheduleAndDoesNotHotLoopOrReleaseResources`. That failure reproduces on pristine `upstream/main` with the same command and is unrelated to this change.

`ArenaTargetFilterUxTest` loads the real `CityEventsExtraLayout.fxml` through `FXMLLoader` and asserts the fields are editable, that each policy enables and disables as its field is filled and cleared, that both stay disabled while arena is off, that each description names the configured tag and state, that a hard skip reads differently from a soft ranking, and that the quoted attempt price matches `ArenaRoutine.extraAttemptGemPrices()`.

Not performed: no live emulator run of the arena routine on this branch, and no screenshot is attached. The change was exercised by hand on a packaged local build — typing a state number enabled the server policy, and both values persisted to the profile and survived a restart — but that was on a downstream build, so it is reported as manual observation rather than as a check on this branch.

## Contributor certification

## Risks

- Profile persistence now happens from a task panel. `persistCurrentProfile()` builds an `AccountDescriptor` the same way `ProfileManagerActionController#toDescriptor` does and calls `ProfileService.persistAccount`, which broadcasts the change so an open profile editor stays in sync. A failed save restores the stored values into the fields rather than leaving a value the run would not use.
- The alliance and server text formatters duplicate the ones in `EditProfileController`; they are held equal by intent, not by a shared constant.
- `BearTrapTimerDateTimeUxTest` is touched: its per-class toolkit latch and `Platform.exit()` were only correct while exactly one test class needed JavaFX. A second class made whichever ran later fail with "Platform.exit has been called". Both now start through `JavaFxToolkit` and neither exits.
- The tab is taller than before. It sits in a `ScrollPane`, so the added descriptions scroll rather than clip.

## Why the checks show red

```
```

Not specific to this branch — filed as #310. Maintainer branches fail identically, and the LFS pointers involved are the repository's own binaries. No branch can go green until that is fixed.

*Upstream behaviour below was checked against `upstream/main`, not recalled.*

The Arena tab shows **Profile alliance** and **Profile server** as read-only `Not set` labels,
so the two values the target-filter policies depend on cannot be entered from the screen that uses
them. The policy dropdowns stay disabled until a profile reload supplies those values from
elsewhere. Nothing on the tab states what a filter does, what **Extra attempts** costs in gems, or
how the free/paid split on **List refresh** works; `extraAttemptGemPrices()`,
`maxFreeListRefreshes()` and `maxPaidListRefreshes()` do not exist upstream, so those figures are
not available to the UI at all.

### What this PR does

Makes both fields editable and write back to the profile, binds the dropdowns to what is typed
so they enable immediately instead of after a reload, and gives every control a description line
including the gem cost of extra attempts and the free/paid refresh split. `ArenaRoutine` exposes the
three accessors so the screen reads the real figures instead of restating them in the layout.

### The difference

Upstream has the settings but no way to reach them from the tab they govern, and no statement
of what any of them costs.

A setting you cannot set from the screen that uses it is effectively not a setting — the
dropdowns being disabled until an unrelated reload is a dead end with no on-screen explanation. The
costs matter more: **Extra attempts spends gems**, and a control that spends a hard currency without
saying so is one an operator can trigger without understanding the price. Reading the figures from
`ArenaRoutine` rather than restating them in the FXML means the displayed price cannot drift away
from the price actually charged.

Opponent selection, scanning, purchasing and refreshing behaviour are unchanged; this is reachability
and disclosure, not a policy change.

### Verification status

3 commits / 6 files, including `ArenaTargetFilterUxTest` and a shared `JavaFxToolkit` test
helper. Java CI has never run — see below.

### DIFF
--- a/modules/desktop/src/main/java/dev/frostguard/app/panel/city/CityEventsExtraLayoutController.java
+++ b/modules/desktop/src/main/java/dev/frostguard/app/panel/city/CityEventsExtraLayoutController.java
@@ -1,9 +1,12 @@
 package dev.frostguard.app.panel.city;
 
 import dev.frostguard.api.configs.ConfigurationKeyEnum;
+import dev.frostguard.api.domain.AccountDescriptor;
+import dev.frostguard.api.domain.ConfigData;
 import dev.frostguard.app.panel.profile.ProfileAux;
 import dev.frostguard.app.shared.AbstractProfileController;
 import dev.frostguard.app.shared.SettingValidators;
+import dev.frostguard.engine.service.ProfileService;
 import javafx.beans.binding.Bindings;
 import javafx.fxml.FXML;
 import javafx.scene.Node;
@@ -11,10 +14,12 @@ import javafx.scene.control.CheckBox;
 import javafx.scene.control.ComboBox;
 import javafx.scene.control.Label;
 import javafx.scene.control.TextField;
+import javafx.scene.control.TextFormatter;
 import javafx.scene.control.Tooltip;
 
 import java.time.LocalTime;
 import java.util.List;
+import java.util.function.UnaryOperator;
 
 public class CityEventsExtraLayoutController extends AbstractProfileController {
 
@@ -30,7 +35,12 @@ public class CityEventsExtraLayoutController extends AbstractProfileController {
         "Avoid profile alliance",
         "Never attack profile alliance"
     );
-    private static final String PROFILE_VALUE_NOT_SET = "Not set";
+    // The same input rules the profile editor enforces, so a value typed here
+    // cannot differ in shape from one typed there.
+    private static final UnaryOperator<TextFormatter.Change> DIGITS_ONLY = change ->
+        change.getControlNewText().matches("\\d*") ? change : null;
+    private static final UnaryOperator<TextFormatter.Change> ALLIANCE_CODE = change ->
+        change.getControlNewText().length() <= 3 && change.getControlNewText().matches("[A-Za-z0-9]*") ? change : null;
 
     @FXML
     private CheckBox checkBoxDailyVipRewards;
@@ -55,9 +65,9 @@ public class CityEventsExtraLayoutController extends AbstractProfileController {
     @FXML
     private TextField textFieldArenaActivationHour;
     @FXML
-    private Label labelArenaProfileAlliance;
+    private TextField textFieldArenaProfileAlliance;
     @FXML
-    private Label labelArenaProfileServer;
+    private TextField textFieldArenaProfileServer;
     @FXML
     private Label labelArenaTargetingHint;
     @FXML
@@ -68,6 +78,8 @@ public class CityEventsExtraLayoutController extends AbstractProfileController {
     private ComboBox<String> comboBoxArenaServerPolicy;
     @FXML
     private Label labelDateTimeError;
+
+    private ProfileAux currentProfile;
 
     @FXML
     private void initialize() {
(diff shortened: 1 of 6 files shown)


bizulk:
I must confess I try to read it, but and partly because it is late, I just don' understand the use case. that is not a bug to be resolved right ?

Probably we need to document this game section first, and what the  task is expected to cover today.
From there the deviation would be easier to understand.

PR #318, opened 2026-09-05
test(automation): initialize workspaces independently of test order

### LINKED ISSUE #317: test(automation): initialize workspaces independently of test order
## Outcome

Make automation tests initialize and isolate their workspace explicitly so clean CI builds do not depend on JUnit/Surefire test discovery order.

## Problem

The Linux build for #316 fails in existing scheduler tests even though that PR changes only scheduled-autostart documentation, XML, and PowerShell. The affected tests access the singleton persistence stack before the configured test workspace exists. A full local run can pass when an earlier service test indirectly creates the workspace, while CI discovers scheduler tests first and fails.

Related CI run: https://github.com/Shederator/wosbot/actions/runs/33962579874
Discovered while reviewing #316; unrelated to the scope of #315.

## Reproduction

From a clean checkout, or after removing generated module output:

    ./mvnw -pl modules/automation -am clean test -Dtest=TaskQueueProfileCooldownTest -Dsurefire.failIfNoSpecifiedTests=false

Observed result:

- 3 tests run, 2 errors
- SQLite reports that modules/automation/target/test-workspace does not exist
- the failure begins through ProfileService.obtain() and DataStore.getInstance()

The full focused reactor test can pass when other tests run first:

    ./mvnw -pl modules/automation -am test

This confirms an order-dependent test fixture rather than a change introduced by #316.

## Scope

- Initialize the configured test workspace before scheduler tests use singleton persistence services.
- Make affected tests create their own persisted profiles instead of relying on data left by another test.
- Reset or close shared persistence state where needed so tests remain independent.
- Keep production workspace ownership and startup behavior unchanged unless a production invariant is proven incorrect.

## Acceptance criteria

- The isolated clean reproduction command passes.
- TaskQueueProfileCooldownTest, TaskQueueInitializeRecoveryTest, and TaskQueueFailureIncidentTest pass without another test class running first.
- The affected automation module tests pass from clean generated output.
- The full reactor package build passes, subject to unrelated external CI infrastructure such as Git LFS availability.
- Tests do not rely on execution order or state created by another test class.

## Evidence

- Automated local reproduction on Linux with JDK 21: deterministic failure using the isolated clean command above.
- GitHub Actions reproduction: six scheduler-test errors with the same missing SQLite parent directory.

## Dependencies and risks

No product behavior dependency. Shared singleton repositories and services may retain process-wide state, so the fix must avoid introducing cross-test contamination or changing runtime workspace ownership.

bizulk:
## Summary

- initialize the test workspace explicitly in the three scheduler test classes that access persistence
- create a unique persisted profile inside `TaskQueueProfileCooldownTest` instead of depending on data left by another test class

## Why

The automation suite was order-dependent: these scheduler tests passed after service tests had created `modules/automation/target/test-workspace`, but failed when CI executed them first. Each test class now establishes the runtime workspace invariant it relies on and creates its own fixture data.

Closes #317.

## Review structure

- Stack: 1 of 1 (not stacked)
- Parent PR: none
- Child PR: none
- Shared issue: #317
- Dependencies: none
- Merge order: this PR is independent and may merge directly

## Validation

- `./mvnw -pl modules/automation -am clean test -Dtest=TaskQueueProfileCooldownTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw -pl modules/automation -am clean test -Dtest=TaskQueueInitializeRecoveryTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw -pl modules/automation -am clean test -Dtest=TaskQueueFailureIncidentTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw -pl modules/automation -am clean test` (237 automation tests passed)
- `./mvnw package` reached and passed the automation module, then failed in unrelated desktop tests because this environment cannot lock Java user preferences and has no JavaFX toolkit

Evidence level: automated tests. Saved-frame and live-log validation are not applicable to this test-only change.

## Contributor certification

- [x] Every commit includes a `Signed-off-by` trailer matching the commit author (DCO 1.1).

## Risks

Low. The change affects test setup only; production runtime behavior is unchanged.

### DIFF
--- a/modules/automation/src/test/java/dev/frostguard/engine/schedule/TaskQueueFailureIncidentTest.java
+++ b/modules/automation/src/test/java/dev/frostguard/engine/schedule/TaskQueueFailureIncidentTest.java
@@ -5,12 +5,15 @@ import dev.frostguard.api.domain.ActionRequiredIncidentData;
 import dev.frostguard.api.domain.ActionRequiredIncidentState;
 import dev.frostguard.api.domain.AccountDescriptor;
 import dev.frostguard.api.domain.LogMessageData;
+import dev.frostguard.api.runtime.WorkspacePaths;
+import dev.frostguard.api.runtime.WorkspaceSession;
 import dev.frostguard.data.repository.TaskFailureStreakRepository;
 import dev.frostguard.engine.error.StartupCaptureException;
 import dev.frostguard.engine.service.ActionRequiredIncidentService;
 import dev.frostguard.engine.service.LoggingService;
 import dev.frostguard.engine.service.ProfileService;
 import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 
 import java.time.LocalDateTime;
@@ -23,6 +26,11 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 class TaskQueueFailureIncidentTest {
+
+    @BeforeAll
+    static void initializeTestWorkspace() {
+        WorkspaceSession.initializeLayout(WorkspacePaths.current());
+    }
 
     @AfterEach
     void detachLogObserver() {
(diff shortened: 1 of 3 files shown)


bizulk:
CI follow-up:

- Linux CI passed, including the scheduler tests that failed on #316. This confirms the order-dependent workspace fixture failure is fixed by this PR.
- DCO passed.
- The Windows installer check failed before checkout/build in `actions/setup-java@v4`: the workflow pins `21.0.12+8.0`, while the current Temurin catalog exposes `21.0.12+8.0.LTS`. This is unrelated to the test-only diff and should be handled separately because the same pin is also used by a signed release workflow.

Failed Windows run: https://github.com/Shederator/wosbot/actions/runs/33969883610

PR #330, opened 2026-09-10
fix(navigation): recover transient sidebar transitions

elroykanye:
Fixes the sidebar transition failure that can make Intel look unavailable and delay it until the next UTC window.

This carries @CodeLtDave's original `350bfe6` with David Erdős preserved as author, then closes the failure-state gaps I reproduced on the Sep 9 Nightly.

- wait through transient sidebar frames and retry once only from fresh Home/World state
- verify the requested section before every list scan or swipe
- distinguish an unavailable sidebar from a genuinely missing Lighthouse row
- keep Intel retryable instead of scheduling the next window after a navigation failure

Proof:
- 22 focused tests
- 644-test package pass, excluding the existing Windows-only `BoundedProcessRunnerTest` failure
- exact final HEAD `e6e6122` packaged as a Nightly image and run on both 720x1280 MuMu profiles
- both live runs found Daily Intel available, entered Intel through Wilderness, and confirmed an Intel marker

Closes #300
Refs #307

PR #332, opened 2026-09-10
fix(intel): preserve unknown stamina readings

### LINKED ISSUE #331: [Bug]: failed stamina OCR is treated as zero and blocks Intel
### Frostguard version or commit

`3.0.2-nightly.20260909.1` and `e6e6122` (the stamina code is unchanged from current `main` at `50dd188`)

### Environment

- OS: Microsoft Windows 11 Pro
- Emulator: MuMu Player 12, two profiles running concurrently
- Resolution / DPI: 720x1280 / 320 DPI

### Actual behavior

After a restart, stamina tracking has no value for either profile. If the initial profile stamina read exceeds the hard 1.5-second cutoff, `StaminaHelper` returns before OCR and never stores a value.

`StaminaService.getCurrentStamina()` then returns `0` for the missing slot. Intel treats that as a measured zero and stops processing, even when the game shows stamina.

Observed live: one profile had about 20 stamina and the other had a full bar. Both were reported as `0` after their reads timed out.

### Expected behavior

Unknown or unread stamina should not be treated as confirmed zero. The read should retry/fail explicitly, or Intel should preserve an unknown state instead of scheduling from false stamina evidence.

### Steps to reproduce

1. Start Frostguard with autostart so the in-memory stamina cache is empty.
2. Let two active profiles initialize concurrently.
3. Observe the profile stamina read exceed 1.5 seconds and abort.
4. Run Intel.
5. Intel reaches the Intel map, then reports `Current stamina: 0` and stops without deploying a mission.

### Sanitized evidence

```text
09:40:13 WARN Stamina read timeout: stamina button click exceeded 1.5s, aborting
09:41:57 WARN Not enough stamina to process intel. Current stamina: 0. Required: 30.
09:42:11 WARN Stamina read timeout: stamina button click exceeded 1.5s, aborting
09:44:28 WARN Not enough stamina to process intel. Current stamina: 0. Required: 30.
```

The source path is direct:

- `StaminaHelper.updateStaminaFromProfile()` returns on timeout without setting stamina.
- `StaminaService.getCurrentStamina()` maps a missing slot to `0`.
- `IntelligenceRoutine.hasEnoughStaminaFlow()` treats values below 30 as confirmed insufficient stamina.

A prior Intel follow-up already noted that stamina OCR fallbacks affected local stamina accounting and needed separate evidence: #274.

This is separate from the sidebar transition fix in #330.

### Privacy check

- [x] I removed credentials, account identifiers, private messages, and unrelated personal data.

elroykanye:
Closes #331.

Failed stamina reads now stay unknown instead of becoming `0`. Intel retries after one minute instead of scheduling a fake regeneration wait.

A real measured zero still uses the normal stamina wait. Deltas made before the first successful read also stay untrusted.

Tested with:
- 22 focused tests covering failed OCR, real zero, inferred deltas, two profiles, and the actual Intel caller
- full automation/task suites
- a local Nightly build on two 720x1280 profiles

This is separate from #330. DCO signed off.


bizulk:
Hello. 
note : In case of a continous failure we could have delayed more the retrial.
But that is just a remaining risk.

PR #336, opened 2026-09-12
feat(tasks): color profile tabs by enabled state

strawhat1814:
## Summary
- Tasks profile tabs show **green** when the profile is enabled, **grey** when disabled, and **blue** when selected.
- Icon refresh prefers the ProfileService snapshot so disabling a profile no longer stays green from a stale TaskQueue copy.

## Why
Idle enabled profiles previously all looked grey because tab dots followed queue position. Users want a clear enabled/disabled signal in Tasks.

## Validation
- `./mvnw.cmd -pl modules/desktop -am compile` — passed
- Live UI: disable a profile in Profiles, confirm Tasks tab dot turns grey (and blue while that tab is selected)
- Evidence level: compile + code review; live confirmation still on the operator

## Contributor certification

## Risks
- Tab dots no longer encode queue position (running/waiting/idle); that remains visible in the task table status column.


strawhat1814:
image

PR #346, opened 2026-09-14
fix(bear): keep rally session active

### LINKED ISSUE #340: [Bug]: Bear Trap loses rally state and misses rallies during the active event
### Frostguard version

v3.0.2, local build at `3b3896d`

### Environment

- Windows
- MuMu Player
- 720 × 1280 portrait

### What happened

Bear Trap hit a navigation error during preparation. The outer exception handler logged it, ran cleanup, and immediately scheduled the next Bear 48 hours later.

That abandons the current event even though most of its 30-minute active window is still available.

The same run also kept trying to call an own rally after one was already active. Those attempts failed with `Hold Rally button not detected`, then continued to rally joining.

### What should happen

Recoverable Bear errors should not end the whole event. While the active window is still open, Bear should:

1. recover to a known screen;
2. wait briefly;
3. retry the failed phase;
4. stop only when the event window ends, the task is cancelled, or the failure is clearly unrecoverable.

Retries must check current state first so they do not create duplicate rallies. If an own rally is already active, skip the call-rally step and continue with joining/checking rallies.

### Evidence

```text
Issue while Bear Trap execution: Home not found after 10 attempts
Cleaning up Bear Trap state
Planning next run Bear Trap for ... [48 hours later]

Hold Rally button not detected!
Could not start rally (may already be active)
```

The task was manually queued again during the same event and successfully resumed Bear participation, confirming that useful work was still possible inside the window.

- [x] Logs are sanitized and contain no credentials or private messages.

elroykanye:
Fixes #340

This replaces the timer/boolean Bear loop with a session driven by the march queue.

What changed:
- tracks the own-rally slot and relaunches only after it is actually idle
- confirms Deploy by finding the new rally row
- reads the selected rally timer and travel time for the cutoff/wake-up hints
- retries navigation, stale-screen, OCR, full/departed rally, formation, and queue failures without ending the event
- fills every verified free join slot in one pass and skips rejected rally rows
- waits on exact return timers instead of polling every second when all marches are busy
- cancellation no longer requeues Gather/Autojoin or schedules Bear 48 hours later

Tests:
- 9 fake-clock/scripted Bear session tests
- 516 reactor tests passed (46 API, 11 persistence, 6 vision, 273 automation, 180 tasks)
- full desktop package built successfully

The two unrelated baseline failures were excluded: `BoundedProcessRunnerTest` and `IntelCycleSchedulingOrchestrationTest`.

Still to validate:
- one live Bear window on both 720x1280 MuMu profiles

Keeping this draft until that live run is done.


elroykanye:
Tested the combined build during today's Bear window. This still needs work, so I'm keeping it draft.

The bot failed to find the own-rally button four times and kept repeating the same navigation. More importantly, the coordinator retries that phase before reaching joining, so a broken own launch can leave every joining march idle.

I also found:
- join selection prefers the bottom visible plus, and rejected Y positions can refer to a different rally after the list refreshes;
- the green-plus match doesn't verify the row is a Bear rally;
- existing-own adoption takes the first rally march, which could be a joined rally;
- the selected timer is read after the Hold dialog has already closed;
- waiting for the own march to return can delay joining even when other slots are free.

The World rally shortcut is already used in `joinNext`; that's not a missing feature. The fix is to verify its transition and make the refresh cycle reliable: World -> fresh rally list -> top eligible Bear join -> available configured flag -> verified Deploy -> World -> reopen.

I'll revise the driver and coordinator around these findings, add real-screen and realistic-timing tests, and validate the complete event at the next Bear window. Today's manual rally launch is not evidence that the bot works end to end. The current scripted tests don't cover the navigation delays or changing list state that failed live.

elroykanye:
Pushed the first revision from the live-run findings (`4707b43`).

Changes:

- A failed own-rally attempt no longer blocks joining.
- The bot now picks the top verified Bear rally from a fresh frame.
- Rally rows are re-read after every refresh instead of reusing stale positions.
- Own and join formations follow the user's configuration. The own formation is reserved, and join order is preserved.
- Existing rally marches are no longer blindly treated as our own rally.
- Rally duration is read while the timer dialog is still open.
- A returning own rally no longer causes a long wait while join slots are free.

Tests: 520 passed.

I'm keeping this as a draft until we validate the full refresh and deploy loop during the next Bear Hunt.

elroykanye:
Synced this draft with current `main` and pushed the Bear changes from the latest live findings.

This revision now:

- stops launching own rallies at **T-5:30**, then stays in join-only mode until the event ends;
- verifies the configured formation by its yellow selected state instead of assuming the tap worked;
- checks both member room and troop capacity before joining;
- reopens War from World and rescans a fresh frame after every attempt;
- keeps the Bear location as a fast return anchor, with the Alliance route only as fallback;
- treats full, departed, stale, and already-joined rows as recoverable outcomes;
- retries unreadable visible Bear rows instead of mistaking OCR failure for an empty rally list;
- removes the active-loop fixed sleeps so captures and confirmed UI transitions set the pace.

This is stacked on #334 because the Bear scanner uses its immutable-frame and verified post-deploy primitives. The focused Bear/navigation suite and full regression suite are green locally.

I’m keeping the PR in draft until the next scheduled Bear Hunt proves the complete launch/join/return loop live.

elroykanye:
Updated from the live findings and synced with current main.

- Reuses the verified Bear camera anchor; Alliance → Territory → Special Buildings is fallback only.
- Opens War directly from a verified World frame and recovers known child screens with one verified Back.
- Explicitly selects five minutes and requires the green tick before starting an own rally.
- Treats “page not ready” separately from an empty list and requires three complete empty scans in the final window.
- Removes the extra two-second activation delay.

Focused Bear tests pass (25 total across timer, navigation, and coordinator coverage), and the full 10-module package is green. Commit: 4fd0e36.

I am leaving this draft until the next live Bear run confirms the timing targets and full event flow.

bizulk:
@elroykanye  This PR is still in draft state, is this correct  ?

elroykanye:
> @elroykanye This PR is still in draft state, is this correct ?

@bizulk yes, I am still working on this

Issue #128, opened 2026-08-07
fix(fishing): provide a complete bounded event workflow


CodeLtDave:
## Problem

Enabling the Fishing Minigame task does not result in a playable event workflow.

The current routine:

- does not navigate to the Fishing Tournament;
- does not open or start a fishing attempt;
- expects the hook to already be visible;
- aborts when the hook is not detected within ten seconds;
- does not handle attempts, results, rewards, or event completion.

`Test Hook Loop` is a diagnostic capture benchmark and does not start the event either.

The existing predictive fish-tracking algorithm is also expensive: every loop captures the screen, detects the hook, and periodically searches for multiple fish templates. Available ADB capture performance may be insufficient for reliable real-time avoidance.

## Proposed behavior

Implement a complete, bounded Fishing Tournament workflow:

1. detect whether the event is active;
2. navigate to the event;
3. detect available attempts or bait;
4. start a fishing attempt;
5. execute a configured control strategy;
6. detect success, failure, timeout, or the result screen;
7. claim applicable rewards;
8. continue while attempts remain;
9. reschedule appropriately when the event or attempts are unavailable.

## Initial control strategy

Provide a low-cost fallback that does not require detecting and tracking every fish.

The initial strategy should be tested from real recordings, for example:

- keep the hook near one screen edge while descending;
- move toward the center when the observed game phase requires it;
- use only the hook position and coarse game-state detection;
- avoid full-frame searches for every visible fish.

The exact movement must be derived from saved frames or recordings rather than assumed.

## Predictive mode

Treat the existing predictive fish-tracking implementation as experimental until it has real-frame and live-run evidence.

It must not be the only available mode. If retained, expose enough performance information to determine whether capture and detection latency are adequate, and abort safely when the effective frame rate is too low.

## Diagnostics

`Test Hook Loop` should be clearly marked as a developer diagnostic or moved out of normal user-facing task configuration. Enabling it must not imply that Frostguard will enter or play the event.

## Safety

- Every attempt has a maximum duration.
- All detection and movement loops support cancellation and preemption.
- Unknown screens stop the attempt instead of producing an indefinite input loop.
- Do not spend gems, premium currency, or paid bait.
- Do not start another attempt until the previous result is confirmed.

## Acceptance criteria

- [ ] The task navigates from the supported start location into the active Fishing Tournament.
- [ ] Available free attempts can be detected and started.
- [ ] The task handles the complete attempt and result flow.
[...]

PR #348, opened 2026-09-15
fix(fishing): add a bounded rewards-first event workflow

elroykanye:
## Summary

Replace the incomplete fishing loop with a bounded, rewards-first workflow: enter the event from home, verify ordinary bait and cast controls, dodge foreground objects on descent, pursue catches on return, verify the Haul and exit.

Add per-profile catch targets, cast/restart limits, Lantern and Stabilizer controls, required-item handling and a random-corner fallback when blind without a Lantern. Short-haul retries reuse the suspended stage, not replacement bait.

The controller uses serial-pinned Android video/touch sessions, retained OCR and foreground motion tracking instead of matching every fish. The bundled scrcpy server retains its license and attribution. Cancellation and cleanup target owned resources.

## Why

Refs #128. The existing routine did not provide a complete event workflow. Ready for review of the bounded, rewards-first core; remaining live-validation limitations are listed below.

Live testing also exposed and corrected shifted/animated home shortcuts, remembered Club tabs, transient bait OCR, stale cast evidence, an animated Haul heading and paid-stage recovery after reopening from home. Go Fish now has a bounded 15-second HUD wait with no second trigger tap.

## Validation

```powershell
.\mvnw.cmd -Pwindows-nightly -pl modules/tasks -am test '-Dtest=Fishing*Test' '-Dsurefire.failIfNoSpecifiedTests=false' '-Dmaven.test.redirectTestOutputToFile=true' -q
```

Passed: 110 fishing tests on both this branch and the local integration branch, zero failures/errors/skips. Saved anonymized real frames cover entry, cast controls, HUD occlusion, depth flashes, item selections, pause/suspension and animated results. Original focused desktop/transport/OCR verification passed 129 tests; later integrated full testing passed 855 tests before the latest corrections. The latest corrections compile in the required reactor and are bundled in the local Windows EXE; app-image verification and native app/watcher smoke pass without system Java.

Live evidence, separately:

- Four diagnostic controller casts earned 1,090 points; the latest filled 20/20. These were not production EXE end-to-end runs.
- An actual production EXE ordinary cast reached 20/20 and a 360-point Haul, but the old animated-heading guard prevented Exit.
- The next ordinary cast was interrupted by a login to the same account elsewhere, confirmed by the operator. Its paid stage survived; the EXE reopened the event and tapped Go Fish, but the old five-second HUD wait expired. That stage produced 270 points. The interruption was not a demonstrated bot crash.
- A private probe invoked only the packaged production `exitVerifiedHaul` method. It verified Exit and return to the overview: 1,720 total points, 5/10 bait, unchanged item counts. This is live exit-component proof, not an uninterrupted EXE cast-to-exit pass.
- The final actual Windows EXE pass completed from home without manual gameplay: bait verified on two fresh frames, one Normal Cast, return detected, full-haul success accepted, automatic Exit verified and the task completed. Account logs at 03:25:47 local record `Fishing haul exited; cast=1 sufficientDescent=true maximumLineConfirmed=false`; at 03:25:48 the one-cast limit rescheduled the task for 60 minutes. The queue recorded successful completion at 03:25:58. No replacement bait, special items or purchases were used.
- Post-run overview confirms 1,990 points (270 gained), bait 4/10 (exactly one spent), and unchanged special-item and voucher counts. The test instance was stopped after completion. The emulator was reopened only to verify this receipt, not to cast again.

No boosters, vouchers, purchases or replacement bait were used. Latest head `1275bc3` passes Linux tests, Windows installer/smoke CI and DCO.

## Contributor certification

- [x] Every commit includes a matching DCO Signed-off-by trailer.

## Risks

- Positive automatic short-haul retry and the extended Go Fish transition wait still need live checks; the final pass filled its haul without exercising these paths.
- Booster restoration, tutorial variants, deep-water darkness, unattended behavior, resource measurements and Apple Silicon runtime remain unverified.
- Ascent is quantity-first, not species/head/value recognition; foreground objects can include non-collectibles.
- Gem buying, multipliers, Horn/Scanner, daily chest and Treasure Hunt are not implemented. Limits are per task session, not daily quotas.
- Upstream project permissions denied status/size updates for #128; its board entry remains stale.


elroykanye:
Both CI jobs hit the same existing Intel test: it expected the next refresh to be more than an hour away. At 23:37 UTC, the next refresh is only about 23 minutes away.

The follow-up checks the next UTC refresh in both the task and persisted schedule, allowing the scheduler's existing sub-second rounding, and keeps the no-hot-loop / no-resource-release checks. Six fixed-time cases check exact refresh boundaries. No production Intel behavior or workflow checks were changed.

Reproduced the original failure locally, then passed all nine focused Intel tests with:

```powershell
.\mvnw.cmd -pl modules/tasks -am test '-Dtest=IntelCycleSchedulingOrchestrationTest,IntelCyclePolicyTest' '-Dsurefire.failIfNoSpecifiedTests=false' '-Dmaven.test.redirectTestOutputToFile=true' -q
```

elroykanye:
Ran the final Windows EXE from home today. It verified bait, made one Normal Cast, accepted the full haul, exited automatically and hit the one-cast limit. The overview confirms 270 points gained and exactly one bait spent; no items, vouchers or purchases. The earlier reload was caused by logging into the same account elsewhere, not a confirmed bot crash. All 110 fishing tests and Linux/Windows installer-smoke/DCO checks pass. Marked ready for review; the remaining live-test limits are in the PR description.

bizulk:
infortunately my fish event has passed, need to test it a bit before.

